Issue #3, opened 2015-12-09
Implement build process to minimize JS bundles


viddo:
Since the grunt tools from cartodb.js was migrated more or less "as-is" we should be able to just apply the [same grunt tasks](https://github.com/CartoDB/cartodb.js/blob/2cf4513bc6901d16f1d39dc9fdf69e9c069fd1ef/Gruntfile.js#L184-L200) in the same manner here.

PR #14, opened 2015-12-15
Fix cartodb.js dependency

### LINKED ISSUE #4: Implement proper bundle task(s) for cartodb.js dependency
Currently, as a temporary workaround, one needs to [set it up manually](https://github.com/CartoDB/deep-insights.js/blob/master/README.md) for things to work, both JS and CSS). We obviously need some better, automate way to setup the dependency. 

Also, depending on how we want deep-insights.js to be used one solution might be more preferable than another, e.g. build one dependency or keep them in separate files, to be loaded (through separate `<script>` tags) in the browser?

This also required for the CI to work

viddo:
Resolves #4 

This simplifies the setup and solves the dependency issue of cartodb.js, so one is not _required_ to do a `npm link` to get the code here working.

Ideally we would define the dependency using typical semver, but since cartodb.js is not (yet) published on npm I set it up to point to a [Github URL](https://docs.npmjs.com/files/package.json#github-urls) (currently points to commit of https://github.com/CartoDB/cartodb.js/pull/936)

@rochoa also mentioned offline some concerns regarding how to handle 3rd party deps (underscore etc.) that is used both in the cartodb.js but also here. If I got it right the proposed solution was to define them as [optional modules](https://docs.npmjs.com/files/package.json#optionaldependencies), but I think the approach with defining dependencies separately and later adding a [`dedupe`](https://docs.npmjs.com/cli/dedupe) step for minimize/publish bundle #3 makes things simpler. Thoughts?

@alonsogarciapablo can you review since I've assigned you on the other one too?


rochoa:
For now I would focus on using dependencies as this PR is doing. I wouldn't over engineer the solution: don't waste time thinking about how to optionally include dependencies.

alonsogarciapablo:
:+1:

PR #22, opened 2015-12-16
Extract source-map and minimize bundle on grunt build

### LINKED ISSUE #3: Implement build process to minimize JS bundles

viddo:
Resolves #3

`grunt build` will now extract source-maps and minimize (uglify) the source file so its size is reasonable to use IRL again…

@xavijam can you review? is essentially the same as in cartodb.js


viddo:
retest this please

viddo:
retest this please

xavijam:
:+1:, what about the compressed version?

viddo:
The compressed is generated with the changes here ofc, when doing `grunt build`

PR #28, opened 2015-12-17
Verify dependencies

rochoa:
This adds a grunt task to validate shared dependencies with cartodb.js.

In order to keep dependencies synced between two projects this will make test task to fail if dependencies mismatch.

Having dependencies in sync is important to not have different ones bundled into the final dist files. This reduces the size of bundles.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -70,8 +70,7 @@ module.exports = function(grunt) {
       })
       .value()
   );
-  grunt.registerTask('test', baseTasks.concat([
-    'lint',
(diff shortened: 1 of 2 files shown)


rochoa:
cc @viddo

rochoa:
Not sure if I should add some kind of test for this.

viddo:
:+1: Let's give this a try

rochoa:
I'm merging it. I hope this in combination with https://github.com/CartoDB/deep-insights.js/pull/32 will help to keep shared dependencies under control.

PR #30, opened 2015-12-17
Fixes examples and updates cartodb.js

alonsogarciapablo:
I'm basically reverting the viz.json files in the examples to the "old" format and pointing cartodb.js to the latest commit. Some examples are still not working because they are local or pointing to an environment where the map/user/wadus no longer exists.

It's working but both CartoDB.js and deep-insights.js are using a WindshaftClient and making a request to Windshaft. This should only happen once, so we should find a way to make it work.

screen shot 2015-12-17 at 19 18 57

cc: @viddo


viddo:
Looks legit. :+1: The conflicts is probably related to the package.json I did, so once fixing that and tests pass go ahead!

PR #32, opened 2015-12-21
Backbone dedupe

rochoa:
This allows to `npm link cartodb.js` without duplicating `backbone` within the generated bundle when using browserify's `--debug` flag.


rochoa:
cc @viddo

viddo:
:ok_woman:

Issue #6, opened 2015-12-10
Change Torque time-series widget behavior on range selection (2nd iteration)


viddo:
The current behavior when a range is selected in a torque time-series widget (as of #786) is to animate the time-slider within the range. When a range is selected the behavior should be changed to:

Awaits https://github.com/CartoDB/cartodb.js/issues/905

- vis/map:
  - [x] all points within the selected time-series should be visible on the map
  - [x] animation must be stopped
- widget:
  - [x] the time-slider should not be available
  - [x] the controls (play/pause) should not be available
  - [x] instead of time-step info the selected range should be displayed (something like _"selected from `<start-time>` to `<end-time>`"_)
  - [x] a button to clear the selected range should be available
  - [x] histogram size/bins should remain after first render of data, even if other filters (e.g. bbox) and/or range selection is changed

PR #33, opened 2015-12-21
Fix time-series widget to render correct views

viddo:
This fixes the stuff that was implemented according against the mock #6 to work again now that the underlying cartodb.js/torque.js implementation actually got done.

fix-time-series

~~Note that selecting a range provokes a re-instantiation of data still, so the selected range seems to be reset, have not yet encountered where that happens exactly, but that can be fixed separately.~~ found the bug, [commented it out for now](https://github.com/CartoDB/deep-insights.js/commit/80b2388ada2e3e8a7d9dbab1bf4a71e17d455415#diff-c6b5f506b4e37c94f215669100b4c54bR94), will create issue to fix separately :point_right: #34 

@rochoa can you review plz?


rochoa:
:ship:ing :it:

PR #35, opened 2015-12-23
Separate dataviews and widgets concepts (WIP)

viddo:
Related to https://github.com/CartoDB/cartodb.js/issues/968, preparing the code that should live in cartodb.js instead of here. 

TBD:
- [x] PoC add dataview concept in addition which widgets use (populated-places examples used)
- [ ] update remaining examples to use dataviews (torque time-series etc.)
- [ ] extract behavior from dataview models that should be in widget models (e.g. `collapse`) where needed (do not apply to all widgets)
- [ ] not all widgets need an own model but can use the dataview model as-is, make it clearer in the code though when that's the case (variable naming/class docs basically)
- [ ] category widget: separate colors management from the dataview since it's more of a visual concern of the widget than the dataview itself(?)


src/dataviews/dataview-model.js
```diff
@@ -0,0 +1,121 @@
...
+  url: function () {
+    return this.get('url') + '?bbox=' + this.get('boundingBox');
+  },
+
+  initialize: function (attrs, opts) {
+    attrs = attrs || {};
+    opts = opts || {};
+
+    if (!attrs.id) {
+      this.set('id', attrs.type + '-' + this.cid);
```
viddo: Widgets defined in a namedmap (see example above) will have ids pre-defined, otherwise one is generated from the model's cid (only valid for the lifecycle of the webpage though).

viddo: I realize, perhaps we should rely on timestamp, or other random number, to avoid clashes in Windshaft

PR #63, opened 2016-01-04
fix shadows in ipad version

piensaenpixel:
@javierarce could you take a look? we need fix the position and when show and hide in ipad version.

- related https://github.com/CartoDB/deep-insights.js/issues/60


viddo:
bump, should this be taken care of as of CartoAssets? or we just close it?

xavijam:
ping @piensaenpixel and @javierarce, should we close it? should we merge it?

PR #75, opened 2016-01-15
Fix time-series torque widget

### LINKED ISSUE #6: Change Torque time-series widget behavior on range selection (2nd iteration)

### LINKED ISSUE #34: Fix torque time-series widget to update range-filter for selected range
In #33 I've [commented it out](https://github.com/CartoDB/deep-insights.js/commit/80b2388ada2e3e8a7d9dbab1bf4a71e17d455415#diff-c6b5f506b4e37c94f215669100b4c54bR94) since it forces the torque-layer to be rendered in faulty state (restarts animation as single step from the start again). Will need to investigate more in-depth why this happens and find a proper solution.

cc @rochoa

viddo:
Resolves #6 
Resolved #34 
hopefully once-and-for-all.

fix-time-series

- [x] Depends on https://github.com/CartoDB/cartodb.js/pull/1003, once merged I'll update the dependency ref.

@rochoa can you review? cc @alonsogarciapablo

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   ],
   "dependencies": {
     "backbone": "1.2.3",
-    "cartodb.js": "CartoDB/cartodb.js#c923025",
+    "cartodb.js": "CartoDB/cartodb.js#5aa87cf",
     "d3": "3.5.8",
     "jquery": "2.1.4",
     "jstify": "0.12.0",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   ],
   "dependencies": {
     "backbone": "1.2.3",
-    "cartodb.js": "CartoDB/cartodb.js#f0bd380",
+    "cartodb.js": "CartoDB/cartodb.js#c923025",
     "d3": "3.5.8",
     "jquery": "2.1.4",
     "jstify": "0.12.0",
--- a/examples/time-series-widget.html
+++ b/examples/time-series-widget.html
@@ -80,7 +80,7 @@
               "type": "public_map",
               "user_name": "orgowner",
               "maps_api_template": "http://localhost.lan:8181/user/{user}",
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
   ],
   "dependencies": {
     "backbone": "1.2.3",
-    "cartodb.js": "CartoDB/cartodb.js#f0bd380",
+    "cartodb.js": "CartoDB/cartodb.js#5aa87cf",
```
rochoa: Are we still using a fixed version for a reason? I mean, couldn't we use `"CartoDB/cartodb.js#v4"`?

rochoa:
About examples... I wouldn't try to rely in staging anymore, and I would start doing examples with documentation production account.

rochoa:
That should be addressed here: https://github.com/CartoDB/deep-insights.js/issues/80

rochoa:
LGTM. I'm :ship: :it: as my only worry is not a blocker.

package.json
viddo: Considering cartodb.js v4 is not versioned and things are changing fast, we have experienced that this repo failed on occasions due to incompatibilities, so we changed back to using "commit-based" versioning for the time-being, to make sure things are always working.

PR #87, opened 2016-01-26
Create dashboard from new vizJSON v3

### LINKED ISSUE #85: Remove diJSON and assume the new data will be provided vizJSON v3
From https://github.com/CartoDB/cartodb/pull/6458 and offline conversations, agreed to remove the concept of a separate "diJSON" and instead have a new version of vizJSON (v3), where the new concepts are included at the root-level (widgets, datasource, title, desc etc.).

- [x] Change createDashboard to take a vizJSON structure instead
- [x] Use new structure as defined in https://github.com/CartoDB/cartodb/issues/6428
- [x] Update all examples

cc @juanignaciosl @alonsogarciapablo

viddo:
Resolves #85 

Review w/o whitespace to easier grasp what actually changed (append `?w=0` to the URL)

@alonsogarciapablo can you review? there are some todos about the stuff I'm skeptical about.


src/create-dashboard.js
```diff
@@ -63,34 +72,36 @@ module.exports = function (selector, diJSON, visOpts) {
...
         });
-      } else if (Number.isInteger(d.layerIndex)) {
-        layer = vis.map.layers.at(d.layerIndex);
-        if (layer && Number.isInteger(d.subLayerIndex)) {
-          layer = layer.layers.at(d.subLayerIndex);
-        }
+      } else if (Number.isInteger(rawWidgetData.layerIndex)) {
+        // TODO Since namedmap doesn't have ids we need to map in another way, here using index
+        //   should we solve this in another way?
+        layer = vis.map.layers.at(rawWidgetData.layerIndex);
```
viddo: While we could remove the sublayer index refs, a namedmap still requires mapping by index, since the dataset ids are not exposed. @rochoa, do we plan to change this finally, so we could :fire: this else-case and just map by ID like the "default" case above?

src/create-dashboard.js
```diff
@@ -63,34 +72,36 @@ module.exports = function (selector, diJSON, visOpts) {
...
     // Create dataview if there's a definition provided
-    var dataviewModel;
-    var d = widget.dataview;
-    if (d) {
+    var dataviewAttrs = rawWidgetData.options;
+    if (dataviewAttrs) {
+      // TODO not ideal, should have a more maintainable way of mapping
+      dataviewAttrs.type = rawWidgetData.type === 'time-series'
+        ? 'histogram' // Time-series widget is represented by a histogram, so re-map the type
+        : rawWidgetData.type;
```
viddo: @alonsogarciapablo one of the issues with changing the vizJSON data structure, since it no longer is a 1to1 match there needs to be some massaging of the data to map things as expected. Any thoughts on how we could do this better?

src/create-dashboard.js
rochoa: Have in mind that layer id is a concept that does not exist at Maps API level. Maps API is agnostic in that sense because it does not need anything like that internally. That means that you can use an id attribute for the layers and Maps API will not intervene modifying/removing it.

Also: we most likely will implement it as an optional attribute in the near future: https://github.com/CartoDB/Windshaft/issues/409. In that case we might need to impose some kind of validation as not all ids will be valid.

But the key thing here is that existing named maps don't include an id attribute either. So I would say it's that easy to just remove the int index option.

viddo: @alonsogarciapablo proposed the same solution, but the widgets doesn't work when we tried that approach. :( The map renders fine but the widgets/dataviews aren't getting any data.

alonsogarciapablo: @viddo I know why tried before, but I don't understand why it didn't worked now... As @rochoa says, there's no need for Windshaft to be aware of this ids. All we need is a way to get the layer by an attribute defined in the widget config (in the viz.json), so as long as the layer models in the the `map.layers` collection have been given the ids that are specified in the viz.json and widgets use those same `layerId`s, the following should work:


```
 layer = vis.map.layers.at(rawWidgetData.layerId);
```

Perhaps we should do a bit of debugging to see why it didn't work before.

src/create-dashboard.js
```diff
@@ -59,45 +59,47 @@ module.exports = function (selector, diJSON, visOpts) {
...
+
       var layer;
 
       // Find the Layer that the Widget should be created for.
-      // a layerId has top-priority, otherwise it tries with a layerIndex, and even a subLayerIndex (if available)
-      if (d.layerId) {
-        layer = vis.interactiveLayers.find(function (m) {
-          return d.layerId === m.get('id');
+      if (layerId) {
+        layer = vis.map.layers.find(function (m) {
```
alonsogarciapablo: Would this work `layer = vis.map.layers.findWhere({ id: layerId });`?

viddo: Indeed, thx!

alonsogarciapablo:
:+1:

PR #91, opened 2016-01-28
Add flag to decide whether to render the menu

### LINKED ISSUE #89: Add flag to not render info view
The info view (left-side top-menu) should only be rendered when dashboard is rendered standalone, we need a flag to decide whether to render it or not, e.g.

```js
createDashboard('#selector', vizJSON, { noInfoView: true })
```

viddo:
Resolves #89

@xavijam review please


spec/dashboard-menu-view.spec.js
```diff
@@ -1,8 +1,8 @@
 var moment = require('moment');
 var Backbone = require('backbone');
-var DashboardInfoView = require('../src/dashboard-info-view');
+var DashboardMenuView = require('../src/dashboard-menu-view');
 
-describe('../src/dashboard-info-view', function () {
+describe('../src/dashboard-menu-view', function () {
```
xavijam: Why not info? It is informative, not a menu :S.

viddo: @acanimal and I thought it was clearer, since it's more of (top-level) menu/bar of some sort.
That said, I don't have a strong opinion, we could change it back if it's important (or come up with some better name  to represent this). What do you think?

src/create-dashboard.js
```diff
@@ -19,19 +20,26 @@ module.exports = function (selector, vizJSON, opts) {
...
 
-  var dashboardInfoModel = new cdb.core.Model({
+  var model = new cdb.core.Model({
     title: vizJSON.title,
     description: vizJSON.description,
     updatedAt: vizJSON.updated_at,
     userName: vizJSON.user.fullname,
-    userAvatarURL: vizJSON.user.avatar_url
+    userAvatarURL: vizJSON.user.avatar_url,
+    renderMenu: opts.renderMenu
```
xavijam: visible?

spec/dashboard-menu-view.spec.js
xavijam: It is fine, let's not discuss this thing, :+1:.

src/create-dashboard.js
viddo: IMHO no, because when set to false it do not _render_ (or create the view), so _visible_ would be misleading.

xavijam:
Come on! :+1:

src/create-dashboard.js
xavijam: Oooook!

PR #92, opened 2016-01-29
Fix problem related with widget loader view

xavijam:
Seems like model is dataviewModel itself:

https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/widget-view.js#L25

RW: @javierarce


javierarce:
:+1: Nice

PR #93, opened 2016-01-29
Widgets public API

### LINKED ISSUE #90: Define public API to add/remove/update widgets
Define & implement how to:
- [x] access/get a widget already added
- add widgets dynamically
  - [x] semantic API
  - [x] ~~`newFoobarModel` requires attrs to have a structure of `{options: {}}`, which comes from the current vizJSON structure. It's kind of weird having this deep structure when it's not strictly needed. IMHO better with a semantic API and have the service map things internally as needed.~~ create-dashboard translates to the service API 
  - [X] validation of args. *If any required param lacks it fails silently showing a message using `cdb.log`*
- remove widgets dynamically
  - [x] awaits https://github.com/CartoDB/cartodb.js/issues/1024 for `dataview.remove`
  - [x] viewmodel.remove() and remove from collection

Postponed/out-of-scope:
- update widgets dynamically (what should be allowed/disallowed?)
- add documentation (in particular the API of the returned objects!)

viddo:
Resolves #90 

Assuming one has created a `dashboard` (`dashboard = cdb.deepInsights.createDashboard('#selector', vizJSON)`)

From there the public API is:

```js
// to access an existing widget:
widgetModel = dashboard.widgets.get('some-id') // returns a widget model object

// To add a new widget (each has their own arguments, see changeset for details):
widgetModel = dashboard.widgets.newCategoryModel();
widgetModel = dashboard.widgets.newHistogramModel();
widgetModel = dashboard.widgets.newFormulaModel();
widgetModel = dashboard.widgets.newTimeSeriesModel();

// To remove a widget (internally cleans up owned objects and removes widget view)
widgetModel.remove();
```

@xavijam @javierarce can you pair-review? :bow: since you're the first users of this
cc @javisantana @alonsogarciapablo FYI


javisantana:
Let me ask a question. The public API is a factory that creates a model, right? is that model a Backbone Model or it just have `remove` method?

acanimal:
Returns a concrete backbone model.

`dashboard.widgets` is a service/factory responsible to, given a set of params, validate and create a concrete `widgetModel`.

We was also though to put a `dashboard.widgets.remove(someWidgetModel)` method too to remove the widgets from the dashboard, but because the `remove` is needed to be implemented too on the widgetModel, to clean up itself, we finally opt to try this approach: create through the dashboard and remove by itself.

javisantana:
ok, thanks

xavijam:
If a editor widget model has been edited, how should we update the deep-insight widget model?

viddo:
Since the actual requirements for what needs to be updated are not entirely clear yet (until we actually start implementing the forms) we left that out of scope (see issue desc).

That said, since we're returning the actual widget backbone models, one have access need to modify their state.  The caveat with this is that one would have to go to the source here to know what methods/attrs there are. 

We should decide if we want to continue down this path, or hide away internals behind some semantic API on these returned objects. There are pros/cons with either approach.

spec/widgets-service.spec.js
```diff
@@ -0,0 +1,274 @@
...
+
+  describe('.newCategoryModel', function () {
+    describe('when given valid input', function () {
+      beforeEach(function () {
+        var attrs = {
+          title: 'some_title',
+          column: 'my_column',
+          aggregation: 'avg'
+        };
+        this.results = this.widgetsService.newCategoryModel(attrs, this.vis.map.layers.first());
```
alonsogarciapablo: In order to make this test a bit more clear, we could rename `this.results` to `this.widgetModel`.

javisantana:
from my experience is better to go in the backbone way to be fast but if you can encapsulate funcionallity is much better, for example, sometimes is better to have:

```
widget.setBins([......])
```
than having

```
widget.set('bins', [....])
```

It does not take so much work and it helps in the long term

javisantana:
in the other hand you need to know something about the attributes to listen for changes, but I guess that's fine for the moment

xavijam:
Ok, so IMHO, we should merge it :+1:.

javierarce:
:+1: Looks good. Let's see how it works for real today :)

viddo:
I like proposal of continuing with backbone models but add semantic methods when/where it makes sense. Let's keep it in mind going forward. :+1:

PR #95, opened 2016-02-01
fixed cat widget

javisantana:
@xavijam this is the bug when creating category widgets. I didn't change the spec

### DIFF
--- a/src/widgets-service.js
+++ b/src/widgets-service.js
@@ -18,10 +18,11 @@ WidgetsService.prototype.get = function (id) {
 /**
  * @param {Object} attrs
  * @param {String} attrs.title Title rendered on the widget view
- * @param {String} attrs.column Name of column
+ * @param {String} attrs.column Name of column to use to aggregate
  * @param {String} attrs.aggregation Name of aggregation operation to apply to get categories
  *   can be any of ['sum', 'count']. Default is 'count'
- * @param {String} attrs.aggregationColumn Name of column to use for aggregation(?)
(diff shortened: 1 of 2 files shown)


xavijam:
cc @viddo @alonsogarciapablo

xavijam:
cc @viddo @alonsogarciapablo @acanimal

src/widgets-service.js
```diff
@@ -36,7 +36,8 @@ WidgetsService.prototype.newCategoryModel = function (attrs, layer) {
   var dataviewModel = this._dataviews.createCategoryModel(layer, {
     type: 'category',
     column: attrs.column,
```
acanimal: This is a good place to ask: what is the difference between `column` and `aggregationColumn` ?

alonsogarciapablo: I think `column` is the name of the column that is used to get the names of the categories, while `aggregationColumn` is the column that is used to aggregate results by.

xavijam: BTW, as @javisantana pointed, we should change that value name to ```aggregation_column```, just to be consistent if later is it exposed in the viz.json.

viddo: Yes, but let's attack and resolve that separately :point_right: #96

xavijam:
Just using default ```column``` value if ```aggregationColumn``` is not provided.
Happy to merge?

viddo:
:shipit:

xavijam:
:face_with_rolling_eyes:

PR #98, opened 2016-02-02
Post public-widgets-api fixes

### LINKED ISSUE #94: Post-public widgets API fixes
Encountered problems while using the new API from #90 
- [x] models are not created with given id
- [x] layerId => layer_id (awaits https://github.com/CartoDB/cartodb/pull/6498)
- [x] rename `widgets.newFooModel` to `widgets.createFooModel`
  - to be consistent with other service objects (e.g. `vis.dataviews`) and be clear that it's not only returning a _new_ object but actually _creates_ it (with everything that happens under the hood).
- [x] show sidebar view when there's at least one widget model

viddo:
Resolves #94 

@javierarce review plz :wink:

### DIFF
--- a/examples/populated-places-local.html
+++ b/examples/populated-places-local.html
@@ -48,7 +48,7 @@
           {
             "type": "histogram",
             "title": "Population",
-            "layerId": "4d82e8df-f21b-4225-b776-61b1bdffde6c",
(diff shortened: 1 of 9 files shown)


javierarce:
:+1: :sunglasses:

PR #104, opened 2016-02-03
fixed timeseries

javisantana:
timeseries need more data that we were sending.


viddo:
:+1: 

> tests are broken, …

Finally was fixed in the last commit 93e3567, right?

javisantana:
yes, sorry

PR #116, opened 2016-02-04
106 update api

### LINKED ISSUE #106: Create public functions to update widgets view
We need to provide a way to update each widget view by passing an argument containing the data.

For example:
```js
dashboard.widgets.updateCategoryWidget('some-widget-id', data); 
widgetModel.updateView(data);
// data should contain all the necessary attributes for each widget.
```

viddo:
Resolves #106

After conversing offline with @xavijam we agreed to go with following the pattern of adding the public API on the previously returned model (like done for [`widget.remove()`](https://github.com/CartoDB/deep-insights.js/pull/93), i.e. having a `widgetModel` is an obvious prerequisite to be able to update _it_.

E.g. having a category widget:
```js
var catWidget = dashboard.widgets.createCategoryWidget({
  title: 'Testing testing…',
  aggregation: 'count'
})
```

Then one can update it by, probably not too surprisingly:
```js
catWidget.update({
  title: 'Bazinga!',
  collapsed: true,
  aggregation: 'sum'
}) // => true
```

Only white-listed attributes are sent through, so trying to update something that's not valid would return false:
```
catWidget.update({
  whatever: 'bazinga!'
}) // => false
```

@xavijam @javierarce does this work for you?

BTW, this solves the problem in the scope of this project, but changing attributes are not reflected on the actual on the dataviews, e.g. try changing a formula's operation from min to max in the populated example. We'll probably have to review all dataviews and make sure they're reactive on attrs changes, to have this fully working, but that can be resolved separately/later. cc @alonsogarciapablo 

---

TODOS
- [x] implement update
- [x] fix attrs not reflected in UI :point_right: #117 
- [x] ~~fix dataview attrs changes not being reflected on widgets~~, the issue is fixed in https://github.com/CartoDB/cartodb.js/pull/1048, need to update the cartodb.js ref here once that's merged
- [ ] Change camelCased attrs to snake_case (by convention), awaits https://github.com/CartoDB/cartodb.js/pull/1052


javisantana:
why creating a new method that does exactly the same than Backbone.Model.set?

For me this PR does not make any sense without adding the dataviews bindings

viddo:
It does not do the same:
- It hides away the dataview so it's not exposed publicly (in the same way as for `widgets.createFooModel` methods).
- It makes sure only attrs that are intended to be updated are set, and on the appropriate model.

javisantana:
My point here is not about functionality, is about flexibility: you just can not create a public api without knowing what you need externally. You only will know what you need when you finish this #117 (which is top priority right now). 

When you are editing models things get weird usually and you need to expose things you wouldn't think in advance. Editing is hard

viddo:
@xavijam updated after fixing #117 as well as https://github.com/CartoDB/cartodb.js/issues/1040, can you review again please?

src/widgets-service.js
```diff
@@ -151,7 +158,9 @@ WidgetsService.prototype.createListModel = function (attrs, layer) {
   var widgetModel = new WidgetModel({
     id: attrs.id,
     type: 'list',
-    title: attrs.title
+    title: attrs.title,
+    attrsNames: ['title'],
+    dataviewModelAttrsNames: ['columns', 'columns_title']
```
xavijam: `columns_title`?

src/widgets-service.js
```diff
@@ -185,7 +194,8 @@ WidgetsService.prototype.createTimeSeriesModel = function (attrs, layer) {
 
   var widgetModel = new WidgetModel({
     id: attrs.id,
-    type: 'time-series'
+    type: 'time-series',
+    dataviewModelAttrsNames: ['column', 'columnType', 'bins', 'start', 'end']
```
xavijam: `column_type` or `columnType`? we have a problem with camelCase and underscore var names :S.

src/widgets/list/content-view.js
```diff
@@ -34,6 +42,11 @@ module.exports = WidgetContentView.extend({
     return this;
   },
 
+  _initBinds: function () {
+    this.listenTo(this.model, 'change:title', this.render);
+    this.listenTo(this._dataviewModel, 'change:data', this.render);
```
xavijam: `this.add_related_model(this._dataviewModel);` maybe ?

xavijam:
Code :+1:

src/widgets-service.js
viddo: I added it since it was included in the examples, it's not used? I'm more than happy to :fire: it is that's the case

src/widgets-service.js
xavijam: The question is, what does it do?

src/widgets-service.js
xavijam: The question is, what does it do?

src/widgets-service.js
viddo: It's due to the [cartodb.js](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/histogram-dataview-model.js#L10) code, I'll send a PR there to update the camelCased ones to have this fixed then so we're consistent.

src/widgets-service.js
xavijam: Ok, it is related with list widget. Let's leave it for the moment, but we will need to review it.

src/widgets/list/content-view.js
viddo: Fixed in d4a2a22 (comment do still appear since the fix was on the next line :()

PR #123, opened 2016-02-05
category none fix

javisantana:
cc @xavijam


xavijam:
Definitely there is no ```NONE``` button anymore.

xavijam:
If my question is yes, code looks :+1:.

javisantana:
that's it

PR #129, opened 2016-02-05
Don't format non numbers

javierarce:


javierarce:
cc: @xavijam

xavijam:
:+1: You are the best,

PR #132, opened 2016-02-05
Updating category items list names to the correct size and make them use Typography styles on cartoassets

saleiva:
cc/ @piensaenpixel @xavijam


xavijam:
:+1:, fixes https://github.com/CartoDB/deep-insights.js/issues/124?

xavijam:
We should "migrate" rest of them, so -> https://github.com/CartoDB/deep-insights.js/issues/133.

saleiva:
fixes #124

PR #138, opened 2016-02-08
Allow override of the default sync settings

### LINKED ISSUE #107: Don't catch faulty errors
https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets-service.js#L166-L171

`cdb.log.error` logs the error but it do not stop execution(!), which means that the creation of widgets return nothing and continue, even though the object creation failed. It should only be used when code execution is safe to continue, this is not one of those cases. We should remove those try-catch'es and just let the error bubble up. This is related the issues found on #104

As a 2nd side note, essentially all these checks are for attrs that belong to the _dataview_ and not the widget, so they should really be verified in cartodb.js instead of here. A dataview should not be created if it lacks compulsory attrs.

cc @acanimal @alonsogarciapablo @javisantana

viddo:
Allows to set `enabled`, `sync_on_data_change` and `sync_on_bbox_change` attrs, to control widgets behavior.

Also, piggy back for other minor fixes that are required for new editor:
- [x] Fixes #107
- [x] ~~Awaits https://github.com/CartoDB/cartodb.js/pull/1054~~ updated dependency
- [x] category dataview, to be reactive to prefix/suffix change, awaits https://github.com/CartoDB/cartodb.js/pull/1056
- [x] push dataviewModel selection of attrs down to dataview instead of here, awaits https://github.com/CartoDB/cartodb.js/pull/1056

@alonsogarciapablo related to the prev PR so you can probably review here too
cc @xavijam @javierarce

### DIFF
--- a/src/widgets-service.js
+++ b/src/widgets-service.js
@@ -3,6 +3,12 @@ var WidgetModel = require('./widgets/widget-model');
 var CategoryWidgetModel = require('./widgets/category/category-widget-model');
 var HistogramWidgetModel = require('./widgets/histogram/histogram-widget-model');
 
+var DEFAULT_DATAVIEW_ATTR_NAMES = [
+  'sync_on_data_change',
+  'sync_on_bbox_change',
+  'enabled'
+];
+
 /**
  * Public API to interact with dashboard widgets.
  */
@@ -47,13 +53,15 @@ WidgetsService.prototype.createCategoryModel = function (attrs, layer) {
     type: 'category',
     title: attrs.title,
     attrsNames: ['title', 'collapsed'],
-    dataviewModelAttrsNames: this._dataviewModelAttrsNames([
-      'column',
-      'aggregation',
-      'aggregation_column',
-      'prefix',
-      'suffix'
-    ])
+    dataviewAttrsNames:
+      [
+        'column',
+        'aggregation',
+        'aggregation_column',
+        'prefix',
+        'suffix'
+      ]
+      .concat(DEFAULT_DATAVIEW_ATTR_NAMES)
   }, {
     dataviewModel: dataviewModel
   });
@@ -88,10 +96,12 @@ WidgetsService.prototype.createHistogramModel = function (attrs, layer) {
     type: 'histogram',
     title: attrs.title,
     attrsNames: ['title', 'collapsed'],
-    dataviewModelAttrsNames: this._dataviewModelAttrsNames([
-      'column',
-      'bins'
-    ])
+    dataviewAttrsNames:
+      [
+        'column',
+        'bins'
+      ]
+      .concat(DEFAULT_DATAVIEW_ATTR_NAMES)
   }, {
     dataviewModel: dataviewModel
   });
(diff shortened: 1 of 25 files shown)


src/widgets-service.js
```diff
@@ -206,10 +222,35 @@ WidgetsService.prototype.createTimeSeriesModel = function (attrs, layer) {
...
+    'sync_on_data_change',
+    'sync_on_bbox_change',
+    'enabled'
+  ].concat(customAttrsNames);
+};
+
+WidgetsService.prototype._dataviewModelAttrs = function (allAttrs, customAttrs) {
+  return _.extend(
+    {
+      sync_on_data_change: _.isBoolean(allAttrs.sync_on_data_change)
```
alonsogarciapablo: Won't this be handled by the dataview itself? (it has default values for these attributes).

src/widgets-service.js
```diff
@@ -31,25 +30,30 @@ WidgetsService.prototype.createCategoryModel = function (attrs, layer) {
...
+    suffix: attrs.suffix
+  }));
 
   var widgetModel = new CategoryWidgetModel({
     id: attrs.id,
     type: 'category',
     title: attrs.title,
     attrsNames: ['title', 'collapsed'],
-    dataviewModelAttrsNames: ['column', 'aggregation', 'aggregation_column']
+    dataviewModelAttrsNames: this._dataviewModelAttrsNames([
```
alonsogarciapablo: I personally find `this._dataviewModelAttrsNames` a bit confusing (name doesn't reveal intention pretty well).

What about being a bit more explicit here?
```
...
dataviewModelAttrsNames: [
  'column',
  ...
].concat(this._defaultModelAttrsNames());
...
```

alonsogarciapablo: We could even use a constant, since these are actually constant:

```
DEFAULT_DATAVIEW_MODEL_ATTRIBUTES = [
  'sync_on_data_change',
  'sync_on_bbox_change',
  'enabled'
]
```

viddo: :+1:

src/widgets-service.js
viddo: Hmm, yes you're right. Probably need another way to pass them through (if they're provided), I'll try to find another way of solving this.

viddo:
@alonsogarciapablo please review the last changes, moved some logic down to cartodb.js since it didn't belong here in the first place, see https://github.com/CartoDB/cartodb.js/pull/1056

src/widgets/widget-model.js
```diff
@@ -26,7 +26,7 @@ module.exports = cdb.core.Model.extend({
   update: function (changes) {
     var attrs = _.pick(changes, this.get('attrsNames'));
     this.set(attrs);
-    attrs = _.pick(changes, this.get('dataviewModelAttrsNames'));
+    attrs = _.pick(changes, this.dataviewModel.constructor.ATTRS_NAMES);
```
viddo: I just realized… we could actually do the same kind of solution on the dataview models, i.e. have a `dataviewModel.update(attrs)` method, that only picks the attrs that it needs and discard the rest. This way  `dataviewModel.constructor.ATTRS_NAMES` can remain private to cartodb.js

viddo:
@xavijam since @alonsogarciapablo seems to busy with the vector integration, could you review this , so we can move forward on the form stuff? Includes checking https://github.com/CartoDB/cartodb.js/pull/1056 to where I extracted the logic from here.

In case there's some disagreement or suggestion for improvement once merged, we can always open new issue(s).

viddo:
@acanimal FYI, since this probably affect carto1b at one point or another, especially the renamed variables (to be `snake_cased`, e.g. `aggregation_column`) + the move of prefix+suffix to widget model.

PR #151, opened 2016-02-10
Fixing problems related with category filters and search state

### LINKED ISSUE #134: Bug when having a single category and clicking on ALL
When having just one category, if you click on it, the ALL button is being shown, but if you click it, it doesn't dissapear even if you are viewing all (which is 1 now)

### LINKED ISSUE #139: Boolean widget (or category widget with 2 options) interaction is wrong
Just on this widget, when selecting one, the other one dissapears and it shouldn't. As it's happening in the Telco PoC

xavijam:
- Displaying the correct info about the selected items, just checking if applied filters are available in the current data.

Before:
screen shot 2016-02-10 at 12 02 42

- Selecting categories already filtered in the search view.

Before:
screen shot 2016-02-10 at 12 04 11

Now:
screen shot 2016-02-10 at 12 05 38

CR: @alonsogarciapablo 
cc @viddo @acanimal
Fixes #139.
Fixes #134.


src/widgets/category/category-widget-model.js
```diff
@@ -136,10 +136,45 @@ module.exports = WidgetModel.extend({
     }
   },
 
+  sizeAcceptedCatsInData: function () {
```
xavijam: I'm on doubt of these methods, should they be here or in CartoDB.js?

@viddo @alonsogarciapablo

xavijam: I'm on doubt of these methods, should they be here or in CartoDB.js?

@viddo @alonsogarciapablo

spec/widgets/category/category-widget-model.spec.js
```diff
@@ -32,6 +33,23 @@ describe('widgets/category/category-widget-model', function () {
     });
   });
 
+  describe('search state', function () {
```
alonsogarciapablo: It looks like we're not really testing the CategoryWidgetModel in these tests. Instead, it looks like are we're testing is the CategoryDataviewModel, so these should probably be moved to CartoDB.js.

spec/widgets/category/options-view.spec.js
```diff
@@ -132,26 +119,28 @@ describe('widgets/category/options-view', function () {
       expect(this.view.$('.js-all').length).toBe(1);
       expect(this.view.$('p.CDB-Widget-textSmaller').text()).toContain('None selected');
     });
+
+    it('should not render all button when there are fewer than 2 categories', function () {
```
alonsogarciapablo: `... when there are less than ...`

spec/widgets/category/options-view.spec.js
```diff
@@ -109,17 +105,8 @@ describe('widgets/category/options-view', function () {
...
-        options.success({
-          'categories': [
-            {category: 'Hey'},
-            {category: 'Buddy'}
-          ]
-        });
-      };
-      this.dataviewModel.fetch();
-      this.view.render();
+      this.dataviewModel._data.reset([{ name: 'hey' }, { name: 'buddy' }, { name: 'neno' }]);
```
alonsogarciapablo: Would something like `spyOn(this.dataviewModel, 'getData').and.returnValue([...])` work here? (just in case the dataview model changes internally and this breaks)

src/widgets/category/category-widget-model.js
alonsogarciapablo: Since we're not really using any data from the widget model itself, they probably belong to CartoDB.js. Deep-insights.js could then use: `dataview.numberOfAcceptedCategories()` and `dataview.numberOfRejectedCategories()`. What do you think?

src/widgets/category/list/items-view.js
```diff
@@ -84,28 +84,29 @@ module.exports = cdb.core.View.extend({
   _setFilters: function (mdl) {
     var isSelected = mdl.get('selected');
     var filter = this.dataviewModel.filter;
+    var clickedName = mdl.get('name');
 
     if (isSelected) {
       if (filter.rejectedCategories.size() === 0 &&
           filter.acceptedCategories.size() === 0 &&
-          this.dataviewModel.getCount() > 1
+          this.dataviewModel.getSize() > 2
```
alonsogarciapablo: Not sure what 2 means by looking at this code. Perhaps we can give that number a name by declaring a constant (`MAX_SOMETHING`)?

spec/widgets/category/category-widget-model.spec.js
xavijam: Well, take into account that we are checking that [this change](https://github.com/CartoDB/deep-insights.js/pull/151/files#diff-9174d99f35e6d799c1c69c256b44b3c3R126) is working, and it is in the `category-widget-model.js`.

src/widgets/category/category-widget-model.js
xavijam: For me makes sense.

spec/widgets/category/options-view.spec.js
xavijam: Thanks :)

spec/widgets/category/options-view.spec.js
xavijam: Well, kind of, getData returns a `Backbone.Collection`, changed.

src/widgets/category/category-widget-model.js
xavijam: https://github.com/CartoDB/cartodb.js/tree/dataview-filter-methods

viddo: :+1: https://github.com/CartoDB/cartodb.js/compare/v4...dataview-filter-methods to make it easier to see the changes :kissing_heart:

alonsogarciapablo: :+1:

spec/widgets/category/category-widget-model.spec.js
alonsogarciapablo: I looked at the code that we are testing, and I'm now wandering If that code actually belongs to deep-insights.js:

There's this binding:

```
  this.listenTo(this.dataviewModel, 'change:searchData', this._onDataviewChangeSearchData);
```

And this callback:

```
 _onDataviewChangeSearchData: function () {
    // Update selected state for each search result item based on locked categories
    this.dataviewModel.getSearchResult().each(function (m) {
      var selected = this.dataviewModel.filter.isAccepted(m.get('name'));
      m.set('selected', selected);
    }, this);
  },
```

It's basically doing stuff on the dataview (not using anything from the widget model as far as I can tell), so perhaps it should be moved to CartoDB.js?

cc: @viddo

viddo: Yes, the repetitive `this.dataviewModel.*` usages is a [feature envy](https://sourcemaking.com/refactoring/smells/feature-envy) smell. The only side-effect that seems relevant for DI/the widget is to update the `selected` based on the search result change, somehow (based on the code snippet here).

xavijam: Yep, the change is only related with the `selected` attribute. Should we move it then?

xavijam:
Something is happening here because travis gets stuck, any idea @viddo?
_BTW, tests in local pass correctly._

xavijam:
Yay! passing tests, or at least seems like that, @alonsogarciapablo ready to go?

alonsogarciapablo:
:+1:

PR #155, opened 2016-02-10
Switched enableFilter and disableFilter, which were internally doing …

alonsogarciapablo:
This should be merged after: cartodb/cartodb#1074. Basically, the internal implentation of `categoryDataview.enableFilter` and `categoryDataview.disableFilter` were wrong, so we were using those methods wrongly. If the widget is locked, the dataview should fetch data using it's own filter (filter should be enabled). It now makes sense:

```
 _onLockedChange: function (m, isLocked) {
    if (isLocked) {
      this.dataviewModel.enableFilter();
    } else {
      this.dataviewModel.disableFilter();
    }
  },
```

@xavijam: makes sense?

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "backbone": "1.2.3",
-    "cartodb.js": "CartoDB/cartodb.js#94bb6f3",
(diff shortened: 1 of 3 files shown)


xavijam:
Yep, totally! :dancer:

PR #156, opened 2016-02-11
Fixes usages of dataview._fetch (which was renamed to fetch)

alonsogarciapablo:
We need to do this because after cartodb/cartodb.js#1079.

@xavijam PTAL! Thanks!


xavijam:
Makes sense :)

PR #158, opened 2016-02-11
Adding proper color to torque histogram widget

### LINKED ISSUE #135: Histogram on Time Series widgets should be orange and not green

xavijam:
After some thoughts, I feel the solution of adding a modifier class works ok, because you don't need to add colors in the JS views, only the class, and colors will be defined in the CSS. What do you think? 

screen shot 2016-02-11 at 18 23 28

CR: @javierarce 
cc @piensaenpixel @viddo

Fixes #135

### DIFF
--- a/src/widgets/histogram/chart.js
+++ b/src/widgets/histogram/chart.js
@@ -5,11 +5,12 @@ var cdb = require('cartodb.js');
 var formatter = require('../../formatter');
 
 module.exports = cdb.core.View.extend({
-  defaults: {
+
+  options: {
     // render the chart once the width is set as default, provide false value for this prop to disable this behavior
     // e.g. for "mini" histogram behavior
     showOnWidthChange: true,
-
+    chartBarColorClass: 'CDB-Chart-bar--green',
     labelsMargin: 16, // px
     hasAxisTip: false,
     minimumBarHeight: 2,
@@ -26,8 +27,6 @@ module.exports = cdb.core.View.extend({
 
   initialize: function () {
     if (!_.isNumber(this.options.height)) throw new Error('height is required');
-
-    this.options = _.extend({}, this.defaults, this.options);
 
     _.bindAll(this, '_selectBars', '_adjustBrushHandles', '_onBrushMove', '_onBrushStart', '_onMouseMove', '_onMouseOut');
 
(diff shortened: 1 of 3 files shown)


src/widgets/histogram/chart.js
```diff
@@ -5,11 +5,12 @@ var cdb = require('cartodb.js');
 var formatter = require('../../formatter');
 
 module.exports = cdb.core.View.extend({
-  defaults: {
+
+  options: {
```
xavijam: Replaced by options, the natural way in the Backbone views.

themes/scss/widgets/themes/light.scss
```diff
@@ -217,42 +219,40 @@ $chartMiniHighlight: rgba(#333, 1);
...
-
-  .CDB-Chart-bar.is-highlighted {
+  .CDB-Chart-bar--green.is-highlighted {
     fill: $darkHighlighted;
   }
-
+  .CDB-Chart-bar--orange {
+    fill: $highlightedAlternative;
+  }
+  .CDB-Chart-bar--orange.is-highlighted {
```
viddo: +1 to use modifier, although :-1: to use the color as the name… why not have something semantic, like `CDB-Chart--timeSeries`?

themes/scss/widgets/themes/light.scss
```diff
@@ -217,42 +219,40 @@ $chartMiniHighlight: rgba(#333, 1);
...
   }
 
-
   //Charts
   .extent {
     stroke: $link;
   }
 
-  .CDB-Chart-bar {
+  .CDB-Chart-bar--green {
```
viddo: Isn't the green color the default color of a char bar? Is it necessary to have a modifier?

themes/scss/widgets/themes/light.scss
xavijam: Yeah, I agree, I was on doubt.
```css
  .CDB-Chart-bar--standard {
    fill: $darkHighlighted;
  }
  .CDB-Chart-bar--timeSeries {
    fill: $highlightedAlternative;
  }
```
Sounds good?

viddo: :ok_woman:

themes/scss/widgets/themes/light.scss
xavijam: :+1:

PR #159, opened 2016-02-11
133 fix texts

piensaenpixel:
@xavijam dont merge yet, but could you take a look?


themes/scss/widgets/text-elements.scss
```diff
@@ -1,72 +1,9 @@
...
-  text-transform: uppercase;
-}
-.CDB-Widget-textSmaller--noEllip {
-  text-overflow: inherit;
-  white-space: inherit;
-  overflow: inherit;
-}
-.CDB-Widget-textBigger--maxWidth {
-  max-width: 304px;
+.is-ellipsis {
```
xavijam: For me, this is a helper, and it should go in CartoAssets.

themes/scss/widgets/form-elements.scss
```diff
@@ -7,17 +7,15 @@
 .CDB-Widget-textInput {
   border: none;
   outline: none;
-  font-family: 'Open Sans';
-  font-size: $sFontSize-large;
 }
 .CDB-Widget-textInput::-webkit-input-placeholder {
-  font-weight: $sFontWeight-lighter;
+  font-weight: 300;
```
xavijam: Don't we already have a variable for that?
Also, didn't we say that 300 was too light?

xavijam:
:+1:  go go go, let's move the ellipsis helper later.

PR #171, opened 2016-02-12
Bring operation badget in formula widgets back!

### LINKED ISSUE #161: [Formula widget] Operation badget is not being displayed anymore
I think it's because [we're trying to get the operation from the widget model](https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/formula/content-view.js#L43), but it's actually a property of the dataview model.

cc: @xavijam

alonsogarciapablo:
Fixes #161:

screen shot 2016-02-12 at 17 22 13

@viddo PTAL. Thanks!


viddo:
:ok_hand: :shipit:

PR #172, opened 2016-02-12
fixes time slider with torque to work in the same way than other widgets

javisantana:
this PR needs other pr to be merged:

- https://github.com/CartoDB/torque/pull/255
- https://github.com/CartoDB/cartodb.js/pull/1091

also needs a new version of the tiler with this commit: https://github.com/CartoDB/Windshaft/commit/d7c39419d2f6849e049d8606d466081ccf4a4b98

cc @viddo

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "backbone": "1.2.3",
-    "cartodb.js": "CartoDB/cartodb.js#177ed9b",
+    "cartodb.js": "CartoDB/cartodb.js#3b6dcfd",
     "d3": "3.5.8",
     "jquery": "2.1.4",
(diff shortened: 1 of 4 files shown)


viddo:
Code looks legit, although logic for calculations of the start/end seems to be duplicated (well, triplicated) across both torque, cartodb.js, and now deep-insights, right? Couldn't we find a way to have that to be provided from a single-source-truth (probably torque)? I suspect that these calculations will cause us trouble further down line otherwise.

javisantana:
what is the logic you see triplicated?

viddo:
I was a little too fast on skimming through the code on Fri, so nevermind. Having a 2nd look there are similar calculations but they're not the same. 
So :+1: to merge I guess (once the tests are fixed).

PR #196, opened 2016-02-16
Skip vis map instantiation and request it when dataviews are generated

xavijam:
Basically, this PR is using the code of this [PR](https://github.com/CartoDB/cartodb.js/pull/1104) from CartoDB.js.

cc @viddo @javisantana @acanimal @alonsogarciapablo @javierarce

### DIFF
--- a/src/create-dashboard.js
+++ b/src/create-dashboard.js
@@ -77,9 +77,7 @@ module.exports = function (selector, vizJSON, opts) {
 
   dashboardView.render();
 
-  vis.instantiateMap({
-    invalidateMapSize: true
-  });
+  vis.instantiateMap();
 
   return {
     dashboardView: dashboardView,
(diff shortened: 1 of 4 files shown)


src/create-dashboard.js
```diff
@@ -75,6 +77,10 @@ module.exports = function (selector, vizJSON, opts) {
 
   dashboardView.render();
 
+  vis.instantiateMap({
+    invalidateMapSize: true
```
viddo: This method call doesn't [match the implementation](https://github.com/CartoDB/cartodb.js/pull/1104#discussion-diff-53049654) (an object instead of a boolean).

But even so, like I [commented on the other PR](https://github.com/CartoDB/cartodb.js/pull/1104#discussion-diff-53049654), why having to pass/know about the "invalidate map size" detail here?

xavijam: Ok!

xavijam:
- [ ] Need to point to the latest CartoDB.js commit.

PR #198, opened 2016-02-17
Makes layer names in cartocss consistent

fdansv:
Internally, the tiler doesn't really mind how you call layers in your cartocss, but for now the d3 renderer is a bit more strict as it uses the cartocss parser without previous treatment. This PR makes layer names consistent.

@alonsogarciapablo


alonsogarciapablo:
:+1:

PR #199, opened 2016-02-17
Added method to re-center map to origin position when there is any widget

### LINKED ISSUE #190: Map isn't centered as specified in .json or by code
Looks like the center of the map is displaced up-left due to the dashboard's panels. 

When you set the view through .json sttings, setView(), .panTo()... you don't get what you set

cc @xavijam

xavijam:
It really fixes #190 issue, but needs [these changes](https://github.com/CartoDB/cartodb.js/pull/1106) in CartoDB.js.

CR: @javisantana

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "backbone": "1.2.3",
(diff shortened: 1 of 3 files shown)


xavijam:
- [ ] Point to the last commit from CartoDB.js#v4.

PR #202, opened 2016-02-17
Fixes histogram normalisation bug

javierarce:
This PR introduces a change that allows to calculate and store the original vertical scale (`yScale`), fixing a bug with the representation of data.

Now the `yScale` is calculated once and stored. The calls to retrieve more data are represented based on the original scale. The scale can be updated (using the `updateYScale` method) or reverted to its original values (with `resetYScale`).

CR: @javisantana

### DIFF
--- a/src/widgets/histogram/chart.js
+++ b/src/widgets/histogram/chart.js
@@ -88,6 +88,7 @@ module.exports = cdb.core.View.extend({
     var labelsMargin = this.model.get('showLabels')
       ? this.options.labelsMargin
       : 0;
+
     return this.model.get('height') - m.top - m.bottom - labelsMargin;
   },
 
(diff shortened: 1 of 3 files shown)


src/widgets/histogram/chart.js
```diff
@@ -467,11 +468,31 @@ module.exports = cdb.core.View.extend({
     this._onWindowResize();
   },
 
-  _setupScales: function () {
+  _getXScale: function () {
+    return d3.scale.linear().domain([0, 100]).range([0, this.chartWidth()]);
```
javisantana: why [0, 100] ?

javierarce: Because those are the values of the brush extents: https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/histogram/chart.js#L547

javisantana: got it

javisantana:
:es:

Issue #153, opened 2016-02-10
Update Category widgets with no metadata on Public View


josecruz:
E.g:
<img width="1096" alt="screen shot 2016-02-12 at 12 26 30" src="https://cloud.githubusercontent.com/assets/596509/13005855/35f6f126-d185-11e5-90f8-d799fb6a6722.png">

Related issue: #147 (histogram) #154 (formula)

Issue #147, opened 2016-02-09
Update histogram widget with no metadata on the public view


saleiva:
![screen shot 2016-02-09 at 19 31 07](https://cloud.githubusercontent.com/assets/132149/12926455/c12a0108-cf63-11e5-9c85-a4a49003344a.png)

Issue #146, opened 2016-02-09
Formula widget on the public view shouldn't show metadata information


saleiva:
![screen shot 2016-02-09 at 19 03 32](https://cloud.githubusercontent.com/assets/132149/12925708/e237375c-cf5f-11e5-9d32-8d5de4f31f40.png)

cc/ @xavijam

PR #203, opened 2016-02-17
Hide widgets stats by default

xavijam:
But if you want to show them, just set `show_stats` attribute to true.
I've found it easy to add, did I miss something? Missing any test?

screen shot 2016-02-17 at 16 37 45
---
screen shot 2016-02-17 at 16 37 50
---
screen shot 2016-02-17 at 16 37 56

CR: @viddo

### DIFF
--- a/examples/populated-places.html
+++ b/examples/populated-places.html
@@ -39,7 +39,8 @@
                 "column": "pop_max",
                 "sync_on_data_change": false,
                 "sync_on_bbox_change": false,
-                "operation": "min"
+                "operation": "min",
+                "show_stats": true
               }
             }, {
               "id": "max-population-formula",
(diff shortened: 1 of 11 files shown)


xavijam:
Fixes #153
Fixes #147
Fixes #146

xavijam:
friendly ping! :dancer:

acanimal:
@xavijam @javierarce Here https://github.com/CartoDB/deep-insights.js/pull/205 I'm i solve a couple of issues forcing to refresh the stats and buttons when widget is expanded.

More problably there will be conflicts with that two PR.

xavijam:
I don't think so @acanimal, this only prevents to render the stats html, that's it.

viddo:
:+1:

PR #204, opened 2016-02-17
Set internal histogram dataview attributes silently

alonsogarciapablo:
Fixes CartoDB/cartodb.js#1069.

It's not the most elegant solution to the problem, but it's the quickest solution I could think off.

@acanimal can you please take a look? Thanks!


acanimal:
:+1:

javisantana:
hey, this is the quickest but this is not the right thing to do. 

When you set parameters with `silent: true` because a third component should not be updated what will happen is you will remove it because you will have another component using those parameters and that component will need to be updated.

So it's much better to go to the end of the chain and remove the binding to those attributes. Specially in this case where start, end and bins are attributes that don't require to update the map.

We need to be quick but not the quickest

PR #205, opened 2016-02-17
WIP - Histogram loses state on collapse/expand

### LINKED ISSUE #160: [Histogram widget] lost state after collapsing and expanding the widget
histogram

cc: @xavijam

acanimal:
CR @javierarce 

close #160 

This PR tries to solved all the issues related with histogram state. When the widget is collapsed and later expanded it:

- loses the stats values, 
- the clear/zoom buttons state,
- lose any filter (on main or zoomed chart) 
- forget the fact if the histogram was zoomed or not, that is, if you are in zoomed mode and collapse/expand the zoomed chart is lost.

### DIFF
--- a/src/widgets/histogram/content-view.js
+++ b/src/widgets/histogram/content-view.js
@@ -146,11 +146,6 @@ module.exports = cdb.core.View.extend({
       this._setupBindings();
       this.$el.toggleClass('is-collapsed', !!this.model.get('collapsed'));
       this._initViews();
-
-      if (!this.model.get('collapsed')) {
-        // Trigger events to force update values
-        this.model.trigger('change:avg change:min change:max change:total change:null change:filter_enabled');
-      }
     }
 
     return this;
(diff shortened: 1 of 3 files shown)


src/widgets/histogram/content-view.js
```diff
@@ -139,6 +139,11 @@ module.exports = cdb.core.View.extend({
       this._setupBindings();
       this.$el.toggleClass('is-collapsed', !!this.model.get('collapsed'));
       this._initViews();
+
+      if (!this.model.get('collapsed')) {
```
xavijam: Weird smell if we have to trigger an *several events* in a render function.

xavijam:
IMHO this is not the solution, shouldn't we keep listening changes although the widget is collapsed? I remember @viddo made some changes in formula widget in order to listen changes in any state.

src/widgets/histogram/content-view.js
viddo: Agree, and keen to end up causing infinite loops due to `change:foo` -> render -> trigger change:foo -> …

acanimal: There are no issue, but I agree is not the right way to solve, it simply was the quickest. I will give another look. Thanks !!

xavijam:
Closing since implementation has changed a lot.

PR #212, opened 2016-02-18
Refresh histogram widget when bins attribute is updated

### LINKED ISSUE #166: if you update the number of bins, the old bins remains displayed as a gray shade
cc @abelvm

xavijam:
update-bins

CR: @javierarce 
cc @viddo in case I miss something for time histogram.

Fixes #166.

### DIFF
--- a/src/widgets/histogram/content-view.js
+++ b/src/widgets/histogram/content-view.js
@@ -93,6 +93,10 @@ module.exports = cdb.core.View.extend({
       return;
     }
 
+    if (this._dataviewModel.changed.bins) {
+      this.binsCount = this._dataviewModel.get('bins');
+    }
+
     // if the action was initiated by the user
     // don't replace the stored data
     if (this.lockedByUser) {
@@ -103,6 +107,7 @@ module.exports = cdb.core.View.extend({
       } else {
         this.histogramChartView.showShadowBars();
         this.originalData = this._dataviewModel.getData();
+        this.miniHistogramChartView.replaceData(this.originalData);
       }
       this.histogramChartView.replaceData(this._dataviewModel.getData());
     }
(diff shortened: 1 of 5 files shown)


xavijam:
cc @AbelVM and @javisantana

xavijam:
Ping!

javierarce:
:+1: Looks good

src/widgets/histogram/content-view.js
```diff
@@ -93,6 +93,10 @@ module.exports = cdb.core.View.extend({
       return;
     }
 
+    if (this._dataviewModel.changed.bins) {
+      this.binsCount = this._dataviewModel.get('bins');
```
viddo: Why are we storing a private var instead of just using `this._dataviewModel.get('bins')` where it's needed?

xavijam: Good question, I don't know, I suppose there is a technical reason, @javierarce could you help me?

javierarce: Originally, when the user zoomed out, I used the `this.binsCount` variable to set the value with the initial value (https://github.com/CartoDB/deep-insights.js/blob/e0695807d3c5b915f8795a4bff41539b1ecad15d/src/widgets/histogram/content-view.js#L436)

xavijam:
There is no `binsCount` variable anymore, thanks to @alonsogarciapablo. Also, maybe you Pablo want to take a look.

src/widgets/histogram/content-view.js
```diff
@@ -76,7 +76,6 @@ module.exports = cdb.core.View.extend({
     if (data && data.length > 0) {
```
alonsogarciapablo: @xavijam Would you remove this whole storeBounds method for me? I should have removed it in a previous PR. It can be :fire::fire::fire:

xavijam: At your command!

PR #214, opened 2016-02-18
Fixes histogram zooming

alonsogarciapablo:
There was a race condition and the zoom/clear feature of the histogram widget wasn't working properly when fetching data from GeoJSON tiles. This PR fixes that. I verified that things are still working :ok_hand: on the raster world with these changes.

histogrrrrram

@javierarce could you please CR? Thanks!

Fixes CartoDB/cartodb.js#1089


javierarce:
:+1: LGTM

PR #220, opened 2016-02-18
Histogram widget view enables and disables histogram dataview filters

alonsogarciapablo:
This should be merged after CartoDB/cartodb.js#1114.

cc: @javisantana

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "backbone": "1.2.3",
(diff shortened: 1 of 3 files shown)


javisantana:
I prefer this approach but it's still a tricky, we are assuming that the mini histogram is the one with the own_filter disabled. That works because is the way it works in the UI.

For me the right thing to do given that the dataview saves state would be to have one data provider per histogram (big and mini)

Anyways, the solution is clean, so good for me

PR #222, opened 2016-02-19
Use append to attach title content so it can be updated

### LINKED ISSUE #221: Updating histogram widget title does not updated the view

acanimal:
CR @javierarce 

close #221 

Title view is new appended to the parent view so it can be updated.

cc @xavijam


javisantana:
could you add some tests please?

acanimal:
Updated code so the widgets shows title although no data was loaded and adde test.

src/widgets/histogram/content-view.js
```diff
@@ -134,6 +138,7 @@ module.exports = cdb.core.View.extend({
 
     if (isDataEmpty) {
       this._addPlaceholder();
+      this._initTitleView();
```
xavijam: Why don't we add the title in the placeholder template? It could save us an unnecessary Backbone view.

acanimal: I suppose initially was on different classes to divede&conquer, but really the title view is really doing a few things. I will change.

acanimal:
Merge title template and view into main view `content-view`.

xavijam:
Sorry but I don't understand why we move all the title logic to the histogram content. I only said that adding a title to the placeholder template was easier.

acanimal:
Then I misunderstood you :( Let me some minutes to undo.

acanimal:
@xavijam In the initial version the main view template has a title but the title subview was "attached" using `.html()` instead `.append()` which makes the title not updatable when the model's title change (really the subview is updated but never shows in the main view because the `html()`).

The reason for that initial title in the main view template is to show the title while the data is loaded, because once loaded the title template is changed by a more complex one used by the `histogra-title-view`, which has some values that are only available when dataview has loaded data.

spec/widgets/histogram/content-view.spec.js
```diff
@@ -12,7 +12,8 @@ describe('widgets/histogram/content-view', function () {
     });
 
     this.widgetModel = new HistogramWidgetModel({
-      title: 'Howdy'
+      title: 'Howdy',
+      attrsNames: ['title']
```
xavijam: What is that?

src/widgets/histogram/histogram-title-template.tpl
```diff
@@ -1,16 +1,18 @@
...
-    <i class="CDB-IconFont CDB-IconFont-drop CDB-IconFont--small CDB-IconFont--top"></i>
-  </button>
-  <button class="CDB-Shape js-actions u-lSpace">
-    <div class="CDB-Shape-threePoints is-blue is-small">
-      <div class="CDB-Shape-threePointsItem"></div>
-      <div class="CDB-Shape-threePointsItem"></div>
-      <div class="CDB-Shape-threePointsItem"></div>
-    </div>
-  </button>
+<div class="CDB-Widget-title CDB-Widget-contentSpaced">
```
xavijam: This template shouldn't change

spec/widgets/histogram/content-view.spec.js
acanimal: Nice question. I think is something we need to change. 

`attrsNames` contains the attributes allowed to be changed. I think this is something good but should be defined as an internal property on each widgetModel instead to be setted on widgetModel createion. (See https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets-service.js#L62)

src/widgets/histogram/histogram-title-template.tpl
acanimal: I wrap into that element because initially it was wrapper on it. See: https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/histogram/content.tpl#L3

spec/widgets/histogram/content-view.spec.js
xavijam: But we don't need that here, attrsNames of the proper HistogramWidgetModel should contain the title, shouldn't it?

src/widgets/histogram/histogram-title-template.tpl
xavijam: You are right, perfect, :+1:.

src/widgets/histogram/content-view.js
xavijam: It is fine for the moment.

xavijam:
Loading state works as it used to?

spec/widgets/histogram/content-view.spec.js
acanimal: I must put otherwise the `this.widgetModel.update()` (see https://github.com/CartoDB/deep-insights.js/blob/221-update-hist-title/spec/widgets/histogram/content-view.spec.js#L112) do not update the title.

acanimal:
I do not modify nothing related with loading state, so it should work fine :)

xavijam:
Check that title appears when histogram is loading and that's it, pretty please :)

acanimal:
Sure, that works. Title is shown although data is loading.

xavijam:
:+1:

PR #223, opened 2016-02-19
fix margins

piensaenpixel:
@xavijam after remove metada in public view we need fix some margins, could you take a look?

### DIFF
--- a/themes/scss/widgets/filter.scss
+++ b/themes/scss/widgets/filter.scss
@@ -4,7 +4,7 @@
 .CDB-Widget-filter {
   display: block;
   max-height: 200px;
-  margin: 10px 0 8px;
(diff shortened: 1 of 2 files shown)


themes/scss/widgets/filter.scss
```diff
@@ -4,7 +4,7 @@
 .CDB-Widget-filter {
   display: block;
   max-height: 200px;
-  margin: 5px 0 $baseSize;
```
xavijam: no basesize?

piensaenpixel: you're right

xavijam:
:+1:

themes/scss/widgets/filter.scss
piensaenpixel: Fixed!

PR #231, opened 2016-02-22
Fix shadow bars problem

### LINKED ISSUE #229: Shadow bars are not working properly

xavijam:
Basically I didn't take into account `originalData` for the shadow bars, and these were using the same data that the current bars have.
I've tried to not "touch" how things are right now, the main change is that `originalData` is now a `Backbone.Collection`, which provides the data to the 'shadowBars'.
If there is any *bins* change, `originalData` would be emptied and then filled with the new data.

CR: @javierarce 
Fixes #229.

### DIFF
--- a/spec/widgets/histogram/content-view.spec.js
+++ b/spec/widgets/histogram/content-view.spec.js
@@ -32,6 +32,13 @@ describe('widgets/histogram/content-view', function () {
     expect(this.view.$('h3').text()).toBe('Howdy');
   });
 
+  it('should reset original data when render the histogram and there is data', function () {
+    spyOn(this.view._originalData, 'reset');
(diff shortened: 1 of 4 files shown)


javierarce:
I found a problem with this branch. Whenever the mini histogram range is changed, this error appears:

```Error accessing array bounds when moving the mini histogram```

screen shot 2016-02-22 at 12 09 21

xavijam:
Taking a look!

xavijam:
Thanks @javierarce, already fixed! You are :top:.

javierarce:
Great! Other than that, the code is :ok_hand:

xavijam:
:clap:

PR #239, opened 2016-02-23
Fixing problem with selected data in histogram, after other widget is filtered

### LINKED ISSUE #235: When other widget filters the data, histogram "selected" stat is not being updated
In the image below, data is filtered in a histogram and selected data is not been reflected in the UI.

screen shot 2016-02-22 at 16 10 50

cc @saleiva

xavijam:
If a widget is filtered, and another histogram belongs to the same layer, stats should be updated.
Also, if bins attribute is updated, stats should be reset too.

CR: @javierarce 
Fixes #235.


javierarce:
:+1:

PR #241, opened 2016-02-23
Fixes creating a time-series model dynamically doesn't render the widget 

### LINKED ISSUE #240: Creating a time-series model dynamically doesn't render the widget
It only renders if it exist in a vizJSON when the dashboard is created up-front, not adding later. Probably missing some event update on some view.

viddo:
Fixes #240 , by changing the visibility on widgets add/removal 

@xavijam @javierarce can any of you review?


xavijam:
:+1:

PR #242, opened 2016-02-23
Fixing problems with the loader and the error state

### LINKED ISSUE #236: Find a way to cancel old requests from widgets
Right now, if a widget is filteres, and then the map is moved, it will trigger two requests (after map instantiation), and it won't cancel the first one, we should find a way to cancel old petitions in order to preserve API performance. Also widget model should check that coming data is from the latest request and not from the old one.

cc @rochoa @alonsogarciapablo

xavijam:
Error state wasn't working, due to a problem with model attribute setting and loader view now is listening to `loaded` event  (https://github.com/CartoDB/cartodb.js/pull/1129/files#r53784263).

CR: @javisantana 

Fixes #236 using https://github.com/CartoDB/cartodb.js/pull/1129/files.

- [x] Update CartoDB.js version (commit pointing to https://github.com/CartoDB/cartodb.js/pull/1129).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "backbone": "1.2.3",
-    "cartodb.js": "CartoDB/cartodb.js#f09747f",
+    "cartodb.js": "CartoDB/cartodb.js#148f575",
     "d3": "3.5.8",
     "jquery": "2.1.4",
     "jstify": "0.12.0",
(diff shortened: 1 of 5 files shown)


javisantana:
:es:

xavijam:
I feel this fix (#236) could be helpful for current examples (density-dot mainly) and probably for you too @AbelVM.

cc @saleiva

PR #243, opened 2016-02-23
Pointing to last CartoDB.js commit

xavijam:
cc @javierarce, nothing to review, but just in case :dancer:


javierarce:
Good to know, thanks!

PR #256, opened 2016-02-25
Fix time slider on range change

viddo:
There was a glitch that until the data returned the time-slider is
visible and animated to the place of the range selector

cc @xavijam


xavijam:
I don't have so much experience with torque widget, but in any case code looks legit so :+1:.

PR #257, opened 2016-02-26
Display HTML in description for embed visualizations

javitonino:
Fixes CartoDB/cartodb#6649
CR @xavijam 
cc @juanignaciosl


xavijam:
Ole! :), :+1:.

PR #261, opened 2016-02-26
Fix 254 Time-series widget renders all data in chart initially instead of only the visible data in bbox

### LINKED ISSUE #254: Time-series widget renders all data in chart initially instead of only the visible data in bbox
Move the map or zoom-in and the correct data is displayed.

Seem to apply to both torque and normal layer.

Probably should do something like how it's done for histogram, i.e. two requests, one for the full data (to have the gray bars behind) and another for the visible data. (w/ or w/o own filter applied I guess).

viddo:
Fixes #254 

Bonus fix: Don't pause the torque animation on data change/range reset :dancer: 

@javierarce review, please :bow:

### DIFF
--- a/src/widgets/histogram/chart.js
+++ b/src/widgets/histogram/chart.js
@@ -314,7 +314,6 @@ module.exports = cdb.core.View.extend({
     this._setupDimensions();
     this._calcBarWidth();
     this._generateChartContent();
-    this._removeShadowBars();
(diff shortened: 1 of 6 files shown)


viddo:
bump :kissing_heart:

javierarce:
:+1: :+1: :+1:

PR #262, opened 2016-02-29
Reset histogram widget when user clicks over CLEAR button

### LINKED ISSUE #258: Clear button is not removed from the view despite of the action is completed
### STR

- Using a histogram widget.
- Select some buckets (filter it).
- Then click over "CLEAR" button.
- After widget has changed with the data, select again some buckets (again filter it).
- Try to click over "CLEAR" button.
- That button doesn't disappear, map is not reloaded and widget is not refreshed.

cc @javisantana

xavijam:
There is a problem with the CLEAR button, and the steps for replicating it are in #258 ticket. We have created a function to reset the state of the histogram widget (origin of the problem). How does it sound @javierarce?

CR: @javierarce.
cc @javisantana 

Fixes #258.

### DIFF
--- a/src/widgets/histogram/content-view.js
+++ b/src/widgets/histogram/content-view.js
@@ -21,7 +21,7 @@ module.exports = cdb.core.View.extend({
   },
 
   events: {
-    'click .js-clear': '_clear',
+    'click .js-clear': '_resetWidget',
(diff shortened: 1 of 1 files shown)


javierarce:
:+1:  LGTM

PR #263, opened 2016-02-29
When a category is selected but it doesn't appear in the current data, "none selected" should appear

xavijam:
Example, selecting BRA category:

- Before (appearing "SELECT ALL"):

screen shot 2016-02-29 at 14 36 40

- Now (appearing "NONE SELECTED"):

screen shot 2016-02-29 at 14 37 16

CR: @viddo

### DIFF
--- a/src/widgets/category/options/options-template.tpl
+++ b/src/widgets/category/options/options-template.tpl
@@ -4,10 +4,10 @@
   <p class="CDB-Text is-semibold CDB-Size-small u-upperCase js-textInfo">
     <% if (isLocked) { %>
       <%- totalCats %> blocked <button class="CDB-Text CDB-Size-small u-upperCase u-actionTextColor CDB-Widget-link u-lSpace js-unlock">unlock</button>
-    <% } else if (areAllRejected || rejectedCats === totalCats) { %>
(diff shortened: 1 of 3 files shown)


src/widgets/category/options/options-template.tpl
```diff
@@ -4,10 +4,10 @@
   <p class="CDB-Text is-semibold CDB-Size-small u-upperCase js-textInfo">
     <% if (isLocked) { %>
       <%- totalCats %> blocked <button class="CDB-Text CDB-Size-small u-upperCase u-actionTextColor CDB-Widget-link u-lSpace js-unlock">unlock</button>
-    <% } else if (areAllRejected || rejectedCats === totalCats) { %>
+    <% } else if (areAllRejected || rejectedCatsInData === totalCats || (acceptedCatsInData === 0 && acceptedCats > 0)) { %>
```
xavijam: We should check that there is any category accepted but there isn't any in the current data.

viddo:
Code wise looks legit, but I have to admit I don't really understand the conditionals in the template. Could the different states be clarified somehow?

xavijam:
Let me try! ;)

viddo:
:+1:

PR #265, opened 2016-03-02
Fixing problem with accepted categories not present in current data

xavijam:
This should fix the problem we have seen today in the last PoC.

screen shot 2016-03-02 at 22 35 08

@saleiva @AbelVM


AbelVM:
Thanks!

PR #268, opened 2016-03-04
Use original data for creating the chart Y scale

xavijam:
Preventing this problem in the zoomed histogram:

Before:
screen shot 2016-03-04 at 11 49 35

After: 
screen shot 2016-03-04 at 11 51 06

cc @javierarce


javierarce:
:+1:

PR #276, opened 2016-03-10
Changing way to use original data within histogram in order to fix 'vector histograms'

### LINKED ISSUE #195: finish time slider for static layers
it does not have for example "clear selection" and so on

### LINKED ISSUE #277: When a torque widget is filtered, widget info is rendered twice
It is related how views are listening changes to renderRange attribute, we should get rid of it because filter model already have the necessary info.

xavijam:
Basically, after [this change](https://github.com/CartoDB/cartodb.js/pull/1181) in CartoDB.js, there is a new way to obtain all the data of the widget. That means we need to change several things in the histogram `content-view` and `chart` classes. I've tried to keep things as they are and only touch affected parts. I'd like to know your opinion @javierarce before keep going.

cc @viddo @javisantana 

Pending things:
- [x] Review tests.
- [x] Add new tests.
- [x] Review the impact over time series/torque widget.
- [x] Review static time-slider widget (fixes #195).
- [x] Review torque info when it is selected (fixes #277).

Overall, fixes https://github.com/CartoDB/cartodb.js/issues/1179.

CR: @javierarce

### DIFF
--- a/src/widgets/histogram/chart.js
+++ b/src/widgets/histogram/chart.js
@@ -65,11 +65,6 @@ module.exports = cdb.core.View.extend({
     return this;
   },
 
-  clean: function () {
-    $(window).unbind('resize', this._onWindowResize);
-    cdb.core.View.prototype.clean.call(this);
-  },
-
   replaceData: function (data) {
     this.model.set({ data: data });
   },
@@ -437,6 +432,7 @@ module.exports = cdb.core.View.extend({
 
   _setupModel: function () {
     this.model = new cdb.core.Model({
+      zoom: false,
       showLabels: true,
       data: this.options.data,
       height: this.options.height,
@@ -498,13 +494,18 @@ module.exports = cdb.core.View.extend({
   },
 
   _setupScales: function () {
+    var data;
     this.updateXScale();
 
     if (!this._originalYScale) {
       this._originalYScale = this.yScale = this._getYScale();
     }
 
-    var data = (this._originalData && this._originalData.getData()) || this.model.get('data');
+    if (this.model.get('zoom')) {
+      data = this.model.get('data');
+    } else {
+      data = this._originalData && this._originalData.getData();
+    }
 
     if (!data || !data.length) {
       return;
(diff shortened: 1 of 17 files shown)


src/widgets/histogram/chart.js
```diff
@@ -497,15 +493,22 @@ module.exports = cdb.core.View.extend({
...
+  _getDataForScales: function () {
+    if (this.model.get('bounded')) {
+      return this.model.get('data');
+    } else {
+      return this._originalData && this._originalData.getData();
+    }
+  },
+
   _setupScales: function () {
+    var data = this._getDataForScales();
```
xavijam: Scale of the histogram depends on the state of itself, if it is bounded (/zoomed) or not.

src/widgets/histogram/content-view.js
```diff
@@ -289,21 +277,12 @@ module.exports = cdb.core.View.extend({
...
-  },
-
   _onChangeFilterEnabled: function () {
     this.$('.js-filter').toggleClass('is-hidden', !this.model.get('filter_enabled'));
   },
 
-  _onChangeBins: function () {
-    this._originalData.reset([]); // Clean originalData
+  _onChangeBins: function (mdl, bins) {
+    this._originalData.setBins(bins);
```
xavijam: This will trigger a new request for getting all the data with that bins.

viddo:
Make sure to verify torque/normal time-series widget still working too, since they also depend on the chart. Just by looking at the code I think there should not be any trouble, but better safe than sorry.

xavijam:
Ha! I didn't finish the pending things in the description, but it was the second point:

screen shot 2016-03-10 at 11 16 05

:scream: 

Thanks @viddo :heart:, now description updated.

xavijam:
About last point in the pending list, self-explanatory gifs:

Before:
torque-bug

Now:
torque-fix

I'd like you @javisantana @viddo  to check the solution, in any case I'd like to know more about the `renderRange` later @javisantana.

xavijam:
So guys, if tests don't play with me, I have this thing ready to be reviewed.
In short:

- Histograms reviewed, the main change using a new model to get all the data.
- Time series widget checked, now it displays "clear" button and selected dates when it is filtered.
- Torque widget "fixed", there was a problem with CartoDB.js and the dates when it was filtered.

@javierarce,  your turn :heart:.

javierarce:
:+1: LGTM

xavijam:
:dancer: :dancers: :dancer: :dancers:

Issue #274, opened 2016-03-09
Long map name doesn't get correctly rendered


saleiva:
https://team.cartodb.com/u/saleiva/editor/dcacc472-e63a-11e5-b837-0e787de82d45/embed

piensaenpixel:
Fixed - https://github.com/CartoDB/deep-insights.js/pull/278

PR #278, opened 2016-03-14
fix long map name

piensaenpixel:
Related - https://github.com/CartoDB/deep-insights.js/issues/274

could you take a look @xavijam ?


xavijam:
:+1:

Issue #275, opened 2016-03-09
Metadata margin is incorrect within the editor


saleiva:
<img width="340" alt="screen shot 2016-03-09 at 16 17 37" src="https://cloud.githubusercontent.com/assets/132149/13650815/7c1049e0-e612-11e5-8611-d34c8a52c673.png">

piensaenpixel:
Fixed - https://github.com/CartoDB/deep-insights.js/pull/279

PR #279, opened 2016-03-14
add margin-top

piensaenpixel:
Related - https://github.com/CartoDB/deep-insights.js/issues/275


xavijam:
Q: Have you considered that stats could not be visible? Margins are ok in that case?

piensaenpixel:
yes, its only when it's visible :)

xavijam:
:+1:

PR #280, opened 2016-03-14
Normalize histogram

javisantana:
adds `normalized` option to histogram.

That means:

- the histogram height will be normalized to the data is currently showing
- the shadow bars are not shown (since they would be in different scales)

cc @xavijam


xavijam:
:+1: :es:, zoomed histogram will be normalized too right?

javisantana:
yes

xavijam:
:us:

xavijam:
From my point of view, all histograms should be normalized by default, as we do with categories.

javisantana:
cc @AbelVM you have this in master :*

PR #288, opened 2016-03-16
Render widgets list sorted by order

xavijam:
And if any widget changes its order, list will be rendered.

CR: @javierarce 
There will be another for CartoDB Editor in minutes, that's for you @viddo :dancer: 

cc @saleiva @AbelVM

### DIFF
--- a/src/dashboard-sidebar-view.js
+++ b/src/dashboard-sidebar-view.js
@@ -49,13 +49,14 @@ module.exports = cdb.core.View.extend({
 
     this._widgets.bind('add', this._maybeRenderWidgetView, this);
     this._widgets.bind('reset', this.render, this);
+    this._widgets.bind('orderChanged', this.render, this);
     this._widgets.bind('change:collapsed', this._onWidgetCollapsed, this);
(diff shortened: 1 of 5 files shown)


src/dashboard-sidebar-view.js
```diff
@@ -49,13 +49,14 @@ module.exports = cdb.core.View.extend({
 
     this._widgets.bind('add', this._maybeRenderWidgetView, this);
     this._widgets.bind('reset', this.render, this);
+    this._widgets.bind('orderChanged', this.render, this);
```
viddo: http://backbonejs.org/#Collection-sort

> Calling sort triggers a "sort" event on the collection.

No need for this custom event

xavijam: Yes, we need it because `sort` event is triggered when a new item is added to the collection. We can't render the list each time a new widget is added in the process :S.

xavijam: http://backbonejs.org/#Collection-sort

viddo: I see, fair enough.

xavijam:
ping @javierarce! :)

javierarce:
:+1:

PR #289, opened 2016-03-16
added switch to enable histogram normalization

### LINKED ISSUE #252: HIstogram normalization hides insights
The image attached shows a gps track of a car in a race track. That zone in the viewport is where the car breaks (really hard). I also added two widgets that show the acceleration (ax and ay). 

When I zoom to that zone I'm expecting to see what's going on in this zone in the histograms but I don't see it because the values are really small compared with the whole dataset.

My point is: we should allow to the user to choose if they want global normalized histograms or not.

screen shot 2016-02-24 at 6 12 32 pm

In the other hand, when the data for an histogram is around 0 we should show a bold marker in the 0 so you see pretty clear where the positive and negative values are.

cc @javierarce @urbanphes

javisantana:
fixes #252 

test

cc @xavijam

### DIFF
--- a/spec/widgets/histogram/chart.spec.js
+++ b/spec/widgets/histogram/chart.spec.js
@@ -77,6 +77,25 @@ describe('widgets/histogram/chart', function () {
     expect(this.view.$el.attr('style')).toMatch('none');
   });
 
+  describe('normalize', function () {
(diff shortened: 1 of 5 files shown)


xavijam:
:+1:

PR #291, opened 2016-03-17
Use main branch for cartodb.js as dependency version

viddo:
Ditto https://github.com/CartoDB/cartodb.js/pull/1199

cc @javisantana


javisantana:
:es:

PR #293, opened 2016-03-18
fix ui

piensaenpixel:
This PR fix this issues

image


xavijam:
In my opinion, the dropdown is too big, how do you like it?

piensaenpixel:
+1 @xavijam but is how it's designed (216px as min-width)

image

maybe @urbanphes could give more information :)

xavijam:
Let's merge it, but we should take a decision about the dimension @urbanphes :heart: :scream:.

PR #296, opened 2016-04-01
Public api draft

javisantana:
this is the proposed public API + the release task.

Please @viddo @xavijam @javierarce @alonsogarciapablo take a look (this needs this PR https://github.com/CartoDB/cartodb.js/pull/1221)

this is the doc: https://github.com/CartoDB/deep-insights.js/blob/api/doc/quickstart.md


doc/quickstart.md
```diff
@@ -0,0 +1,92 @@
...
+
+### cartodb.DI.Dashborbard
+
+This object contains a CartoDB dashboard, a map, and some widgets:
+
+#### getWidgets() -> Array of widgets
+#### getWidget(id) -> get widget
+#### getMap() -> returns a Map object
+
+#### createCategory-> CategoryWidget
```
alonsogarciapablo: IMO `createCategoryWidget`, `createHistogramWidget`, etc. would be more descriptive.

src/api.js
```diff
@@ -0,0 +1,49 @@
+var _ = require('underscore');
+var createDashboard = require('./create-dashboard');
+var cdb = require('cartodb.js');
+
+function Dashboard (dashboard) {
+  this.dashboard = dashboard;
+
+  // bind widget related methods
+  var ws = dashboard.widgets;
+  this.createCategory = ws.createCategoryModel.bind(ws);
```
alonsogarciapablo: I like this API object :+1: 

Since these methods are going to be public (as in API public), I'd be nice to declare them outside of the constructor and document them (something basic):
```
/**
 * Create a category widget.
 * @param {Object} widgetAttrs - attributes for the new widget
 * @param {string} widgetAttrs.id - id (required)
 * @param {string} widgetAttrs.title - title (required)
 * @param {number} widgetAttrs.order - index of the widget (optional)
 * @param ...
 * @return {CategoryWidget} The new widget
 */
Dashboard.prototype.createCategoryWidget = function (widgetAttrs) {
  this._dashboard.widgets.createCategoryModel.call(this, widgetAttrs);
}
```

Perhaps this would help us generate some nice docs?

doc/quickstart.md
javisantana: agree, will change it

src/api.js
javisantana: :+1:

javisantana:
@alonsogarciapablo check again, I changed some stuff based on your comments

doc/api.md
```diff
@@ -0,0 +1,120 @@
...
+#### Arguments
+- domObject: dom object identifier where the dashboard will be inserted
+- vizJson: either a vizjson url or a vizjson object. See viz.json spec to see how to generate one
+- callback: it will be called when the dashboard is generated. First argument will be null if
+  everything was fine, an `Error` object otherwise. Second argument is a `Dashboard` object, see
+  below
+
+#### Example
+```js
+cartodb.deepInsights.createDashboard('#dashboard', vizJSONurl, function(err, dashboard) {
```
alonsogarciapablo: We're missing [the options](https://github.com/CartoDB/deep-insights.js/pull/296/files#diff-e1d62d8c4965edf0e532dfa19da0891fR87) argument here.

doc/api.md
```diff
@@ -0,0 +1,120 @@
...
+});
+```
+
+## cartodb.DI.Dashborbard
+
+This object contains a CartoDB dashboard, a map, and some widgets:
+
+##### getWidgets() -> Array of widgets
+##### getWidget(id) -> get widget
+##### getMap() -> returns a Map object
```
alonsogarciapablo: Why are we providing a method to retrieve the `map` instead of the whole `vis`? If we return the `vis` object we can always refer people to the CartoDB.js docs so that they know what to do with that object.

doc/api.md
```diff
@@ -0,0 +1,120 @@
...
+
+Optional:
+- `id`: identifier that will be used when calling `getWidget`
+- `order`: order inside the widget list
+- `collapsed`: show the wiget collapsed if true
+- `show_stats`: show histogram stats
+- `normalized`: normalize data seen in the map bounding box with the global dataset histogram 
+
+##### createFormulaWidget(widgetAttrs, layer) -> FormulaWidget
+widgetAttrs: `id`, `title`, `order`, `collapsed`, `prefix`, `suffix`, `show_stats`, `description`
```
alonsogarciapablo: Do we want to document these attributes?

doc/api.md
```diff
@@ -0,0 +1,120 @@
...
+
+##### createTimeSeriesWidget(widgetAttrs, layer) -> TimeSeriesWidget
+widgetAttrs: `id`, `title`, `order`, `collapsed`, `bins`, `show_stats`, `normalized`
```
alonsogarciapablo: Same for these.

doc/api.md
javisantana: It's the whole Vis but as we changed that in the editor let's stick to that naming.

javisantana: We will change Vis to Map in cartodb.js as well

alonsogarciapablo: :+1:

src/api.js
```diff
@@ -0,0 +1,114 @@
...
+   * @param {Integer} id - widget id
+   * @return a widget object
+   */
+  getWidget: function (id) {
+    return this.dashboard.widgets.get(id);
+  },
+
+  /**
+   * Create a category widget.
+   * @param {Object} widgetAttrs - attributes for the new widget
```
alonsogarciapablo: It seems that documentations of params for `createCategoryWidget `, `createHistogramWidget`, `createFormulaWidget`, and `createTimeSeriesWidget` are all the same (i.e: not the real ones).

javisantana: Yes, it's not the objetive to generate a nice javadoc (the actual content is valid tho)

javisantana: I mean, it's not the same

doc/api.md
javisantana: I though it would be interesting to have the params there, do you want to remove or you mean we should document them

alonsogarciapablo: It seems that they are documented thoroughly for createHistogramWidget but not so well document for the rest of these methods. You can decide what level of detail is best at this point, but It'd be nice to keep it consistent.

src/api.js
alonsogarciapablo: I'm missing some important attributes here. eg: `column` for the category widget or bins for the histogram widget. It looks like common attributes for all widgets are documented here, but some other important ones are not. That's what I wanted to say with "are all the same".

alonsogarciapablo:
Be aware that this PR breaks the new editor because we're changing the public interface of createDashboard... Take a look at [this usuage](https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb3/editor.js#L37-L41).

PR #298, opened 2016-04-07
Don't take into account layer visibility in order to show widget view

xavijam:
As we decided when we talked about necessary analysis use cases, if a layer is not visible, widgets can be added and they will be visible.
cc @javierarce @alonsogarciapablo @rochoa


alonsogarciapablo:
:+1:

PR #300, opened 2016-04-13
Add toggle in dropdown

piensaenpixel:
@xavijam sorry but could you take a look?

image


xavijam:
Let's fix tests first please :pray:

src/widgets/category/title/search-title-view.js
```diff
@@ -31,42 +31,46 @@ module.exports = cdb.core.View.extend({
...
   _initBinds: function () {
-    this.widgetModel.bind('change:search', this._onSearchToggled, this);
-    this.widgetModel.bind('change:title change:collapsed change:isColorsApplied', this.render, this);
-    this.widgetModel.lockedCategories.bind('change add remove', this.render, this);
-    this.add_related_model(this.widgetModel);
-    this.add_related_model(this.widgetModel.lockedCategories);
+    this.model.bind('change:search', this._onSearchToggled, this);
+    this.model.bind('change:title change:collapsed change:pinned change:isColorsApplied', this.render, this);
+    this.model.lockedCategories.bind('change add remove', this.render, this);
+    this.add_related_model(this.model);
```
xavijam: This is not needed if we use `this.model`.

javierarce: True. Fixed.

piensaenpixel:
ping @xavijam

src/widgets/category/title/search-title-view.js
```diff
@@ -22,7 +22,7 @@ module.exports = cdb.core.View.extend({
   },
 
   initialize: function () {
-    this.widgetModel = this.options.widgetModel;
+    this.model = this.options.widgetModel;
```
xavijam: Why have we changed the name of the variable? just out of curiosity.

javierarce: Because we use `this.model` in the other content-view files.

src/widgets/dropdown/template.tpl
```diff
@@ -1,14 +1,44 @@
 <ul class="CDB-Dropdown-list CDB-Text CDB-Size-medium">
   <li class="CDB-Dropdown-item">
-    <button class="CDB-Dropdown-link js-toggle">Toggle widget</button>
+    <button class="CDB-Dropdown-link js-toggleCollapsed">
+      Toggle widget
+        <div>
+        <% if (collapsed) { %>
```
xavijam: I'd try to not repeat the same html tag, only add in the condition if the input is checked or not.

src/widgets/dropdown/template.tpl
```diff
@@ -1,14 +1,44 @@
...
+          <span class="u-iBlock CDB-ToggleFace"></span>
+        </div>
+    </button>
   </li>
   <li class="CDB-Dropdown-item">
-    <button class="CDB-Dropdown-link js-pin">Pin widget</button>
+    <button class="CDB-Dropdown-link js-togglePinned">
+        Pin widget
+        <div>
+          <% if (pinned) { %>
```
xavijam: Same than above.

src/widgets/dropdown/template.tpl
```diff
@@ -1,14 +1,44 @@
...
   </li>
 
-  <% if (normalizeHistogram) { %>
+  <% if (flags.normalizeHistogram) { %>
   <li class="CDB-Dropdown-item">
-    <button class="CDB-Dropdown-link js-normalize">Normalize</button>
+    <div class="CDB-Dropdown-link js-toggleNormalized">
+      Normalize
+      <div>
+        <% if (normalized) { %>
```
xavijam: Same as above * 2.

src/widgets/dropdown/widget-dropdown-model.js
```diff
@@ -0,0 +1,16 @@
+var cdb = require('cartodb.js');
+
+/**
+ * Default widget dropdown model
+ */
+module.exports = cdb.core.Model.extend({
+  defaults: {
+    collapsed: false,
+    flags: {
```
xavijam: what is the difference between normalizeHistrogram and normalized?

src/widgets/dropdown/template.tpl
```diff
@@ -1,14 +1,44 @@
...
+          <% } else { %>
+            <input class="CDB-Toggle js-inputPinned" type="checkbox" name="pinned">
+          <% } %>
+          <span class="u-iBlock CDB-ToggleFace"></span>
+        </div>
+    </button>
   </li>
 
-  <% if (normalizeHistogram) { %>
+  <% if (flags.normalizeHistogram) { %>
```
xavijam: I'd try to use the necessary variables from the view and not using `var.whatever`.

src/widgets/dropdown/widget-dropdown-view.js
```diff
@@ -87,19 +90,22 @@ module.exports = cdb.core.View.extend({
     }
   },
 
-  _pin: function () {
-    this.model.set('open', false);
-    this.trigger('click', 'pin');
+  _togglePinned: function () {
+    var pinned = !this.model.get('pinned');
```
xavijam: When something is clicked, shouldn't the model "whatever" attribute change?

src/widgets/histogram/content-view.js
```diff
@@ -36,20 +36,28 @@ module.exports = cdb.core.View.extend({
     this._initTitleView();
 
     var dropdown = new DropdownView({
+      normalized: this.model.get('normalized'),
```
xavijam: I'd pass the current model and change the values in the dropdown view. With that we wouldn't need another local model and passing the event to the parent view in order to change de "attribute" value.

src/widgets/dropdown/template.tpl
javierarce: We were getting linting errors when we wrote the condition of the template inside of an HTML tag. Writing something like <div <% if (this) { that } %> class="wait-what"></div> aborted the compilation of the code.

src/widgets/dropdown/widget-dropdown-model.js
javierarce: I'm not sure, it was originally written this way. I didn't want to change the existing behaviour.

src/widgets/dropdown/template.tpl
xavijam: We can use ternaries. They work pretty well in these contexts.
`<%- this === true ? 'paco' : '' %>`

xavijam:
As I have mentioned I feel we could get rid of the dropdown model and keep using the widget model, that's my only concern. About the rest, some questions asked but in general changes are great.

src/widgets/histogram/content-view.js
javierarce: 👌

src/widgets/dropdown/template.tpl
javierarce: I think we tried that and the problem was the same. Inside of an HTML tag (ie: `<div here>`) the template tags would only work if they where inside of some quotes. Otherwise, they'd make the linter fail :(

xavijam: But linter is not enabled for .tpl files, isn't it?

javierarce: Yes, sorry, you are right… I'm not referring to semistandard, when I say "linter" I mean that the code doesn't compile due to an internal validation.

javisantana:
We should do what @xavijam says if we want to be able to  serialize widget state. Do you need help @piensaenpixel ?

javisantana:
cc @xavijam 

hey. I fixed the thing. I'm finding a recurrent pattern with models that contain view states: we are using the model to keep the state and not to manage the state. So we are doing:

user event -> modify the model and change the view 
and it should be

user event -> modify the model -> change the view 
cc @CartoDB/frontend

src/widgets/dropdown/widget-dropdown-view.js
```diff
@@ -93,19 +88,19 @@ module.exports = cdb.core.View.extend({
   _togglePinned: function () {
     var pinned = !this.model.get('pinned');
```
xavijam: Why not render the component when any of those attributes have changed within the model?

xavijam:
Agree @javisantana

src/widgets/dropdown/widget-dropdown-view.js
javisantana: it's being done (check _initBinds ` this.model.bind('change:pinned change:collapsed change:normalized', this.render, this`)

src/widgets/dropdown/widget-dropdown-view.js
```diff
@@ -68,38 +71,41 @@ module.exports = cdb.core.View.extend({
...
+    var pinned = !this.model.get('pinned');
+    this.$('.js-inputPinned').attr('checked', pinned);
+    this.model.set('pinned', pinned);
   },
 
-  _toggle: function () {
-    this.model.set('open', false);
-    this.trigger('click', 'toggle');
+  _toggleCollapsed: function () {
+    var collapsed = !this.model.get('collapsed');
```
xavijam: So we don't need this thing anymore, because render will display checked as it should be.

src/widgets/dropdown/widget-dropdown-view.js
```diff
@@ -68,38 +71,41 @@ module.exports = cdb.core.View.extend({
...
       this._close();
     }
   },
 
-  _pin: function () {
-    this.model.set('open', false);
-    this.trigger('click', 'pin');
+  _togglePinned: function () {
+    var pinned = !this.model.get('pinned');
+    this.$('.js-inputPinned').attr('checked', pinned);
```
xavijam: Same here.

src/widgets/dropdown/widget-dropdown-view.js
```diff
@@ -68,38 +71,41 @@ module.exports = cdb.core.View.extend({
...
+    this.$('.js-inputCollapsed').attr('checked', collapsed);
+    this.model.set('collapsed', collapsed);
   },
 
-  _normalize: function () {
-    this.model.set('normalize', !this.model.get('normalize'));
-    this.trigger('click', 'normalize', this.model.get('normalize'));
+  _toggleNormalized: function () {
+    var normalized = !this.model.get('normalized');
+    this.$('.js-inputNormalized').attr('checked', normalized);
```
xavijam: Same here.

javisantana:
retest this please

src/widgets/category/title/search-title-view.js
fdansv: Isn't this a bit sacrificing clarity for convention, where we don't really need to follow a convention? I find now myself constantly forgetting whether `this.model` refers to the dataview or the widget.

PR #304, opened 2016-04-18
Auto style

fdansv:
* [x] Put CartoCSS generation somewhere else
* [x] Rename apply colors/apply sizes to auto style
* [x] Merge https://github.com/CartoDB/cartodb.js/pull/1232
* [x] Testing

@javisantana @alonsogarciapablo


src/widgets/histogram/histogram-title-view.js
```diff
@@ -50,12 +54,59 @@ module.exports = cdb.core.View.extend({
     this.addView(sizesTooltip);
   },
 
-  _applySizes: function () {
-    this.dataviewModel.set('histogram_sizes', true);
+  autoStyle: function () {
+    var data = this.dataviewModel.get('data');
```
javisantana: this method should be outside of here, it's likely we will need this in other parts.

It should be something like

```
var cartocss = pecanByHistogram(histogram, stylingOtions [, geometryType])
```

src/widgets/category/category-widget-model.js
```diff
@@ -55,16 +53,31 @@ module.exports = WidgetModel.extend({
     this.lockedCategories.reset([]);
   },
 
-  applyColors: function () {
-    this.set('isColorsApplied', true);
+  autoStyle: function () {
+    var style = this.autoStyler.getStyle();
+    if (!this.dataviewModel._dataProvider) {
```
alonsogarciapablo: IMO we shouldn't access `_dataProvider` here... It's supposed be a private thing. My suggestion would be to remove the "if" and just set the new styles on the layer, letting CartoDB.js take care of "everything" else internally. Something like:

```
var style = this.autoStyler.getStyle();
this.dataviewModel.layer.set('cartocss', style);
this.dataviewModel.set('autoStyle', true); // Not sure how this is being used
```

There's already [this binding](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js#L25-L28), wouldn't that update the styles on the "vector sublayer"?

src/widgets/category/category-widget-model.js
```diff
@@ -55,16 +53,31 @@ module.exports = WidgetModel.extend({
...
+      var sublayer = this.dataviewModel._dataProvider._vectorLayerView;
+      sublayer.setCartoCSS(index, style, true);
+    }
+    this.dataviewModel.set('autoStyle', true);
   },
 
-  cancelColors: function () {
-    this.set('isColorsApplied', false);
+  cancelAutoStyle: function () {
+    if (!this.dataviewModel._dataProvider) {
```
alonsogarciapablo: Same here, I'd do something like:

```
this.dataviewModel.layer.restoreCartoCSS();
this.dataviewModel.set('autoStyle', false);
```

src/widgets/auto-styler.js
```diff
@@ -0,0 +1,141 @@
+var cdb = require('cartodb.js');
+var CategoryColors = require('./category/category-colors');
+var AutoStyler = cdb.core.Model.extend({
+  initialize: function (dataviewModel) {
+    this.dataviewModel = dataviewModel;
+    this.colors = new CategoryColors();
+    this.vector = !!this.dataviewModel._dataProvider;
```
alonsogarciapablo: Again, this class shouldn't know about how the map is rendered... That was the goal of deep-insights.js (separating widgets from map rendering). In this class we have several conditionals that make it work differently for raster and vector... CartoDB.js's API should provide methods so that we don't need to things like accessing `this.dataviewModel._dataProvider`...

src/widgets/auto-styler.js
```diff
@@ -0,0 +1,141 @@
...
+    return style
+      .replace(/{{defaultColor}}/g, defColor)
+      .replace('{{markerWidth}}', 10);
+  },
+
+  _getLayerHeader: function (symbol) {
+    return '#' + this.dataviewModel.layer.get('layer_name') + '[mapnik-geometry-type=' + AutoStyler.MAPNIK_MAPPING[symbol] + ']{';
+  },
+
+  _getGeometryType: function () {
```
alonsogarciapablo: For example. if deep-insights.js needs to get the geometry type from a layer, we can implement this in CartoDB.js so that we can just do something like `this.dataviewModel.layer.getGeometryType()` here (and remove the `if (this.vector) { ... } else { ... } `...

src/widgets/auto-styler.js
```diff
@@ -0,0 +1,141 @@
...
+  initialize: function (dataviewModel) {
+    this.dataviewModel = dataviewModel;
+    this.colors = new CategoryColors();
+    this.vector = !!this.dataviewModel._dataProvider;
+  },
+
+  getStyle: function () {
+    var style;
+    var widgetType = this.dataviewModel.get('type');
+    if (widgetType === 'category') {
```
alonsogarciapablo: Just worth mentioning that this could be a good opportunity to [replace this conditional with polymorphism](https://sourcemaking.com/refactoring/replace-conditional-with-polymorphism) (in combination with a factory):

```
var AutoStylerFactory = {
  get: function (dataviewModel) {
    if (dataviewModel.get('type') === 'category') {
      return new CategoryAutoStyler(dataviewModel);
    } else if (dataviewModel.get('type') === 'histogram') {
      return new HistogramAutoStyler(dataviewModel);
    } else {
      throw new Error('dataview type not supported');
    }
  }
}
...
var histogramAutoStyler = AutoStylerFactory.get(histogramDataview);
histogramAutoStyler.getStyle();
...
var CategoryAutoStyler = AutoStylerFactory.get(categoryDataview);
categoryAutoStyler.getStyle();
```

This way widgets could something like:
```
this.autoStyler = AutoStylerFactory.get(this.dataviewModel);
...
this.autoStyler.getStyle();
```

src/widgets/category/category-colors.js
```diff
@@ -12,20 +15,20 @@ var defaultColor = '#CCC';
...
 CategoryColors.prototype.updateData = function (d) {
   // Remove categories from colors where they are not present anymore
-  _.each(this.colors, function (value, key) {
-    if (!_.contains(d, value)) {
-      this.colors[key] = null;
-    } else {
-      d = _.without(d, value);
-    }
-  }, this);
+  // _.each(this.colors, function (value, key) {
```
alonsogarciapablo: There's some commented code here.

src/widgets/category/category-widget-model.js
```diff
@@ -55,16 +53,31 @@ module.exports = WidgetModel.extend({
...
+      var sublayer = this.dataviewModel._dataProvider._vectorLayerView;
+      sublayer.renderers[index].restoreCartoCSS(true);
+    }
+    this.dataviewModel.set('autoStyle', false);
   },
 
-  isColorApplied: function () {
-    return this.get('isColorsApplied');
+  isAutoStyle: function () {
+    return this.dataviewModel.get('autoStyle');
```
alonsogarciapablo: Could "autoStyle" be a property of the widgetModel? What do you think about using `autoStyled` (d at the end) and `isAutoStyled`?

src/widgets/auto-styler.js
alonsogarciapablo: If you follow this route... perhaps we can merge `CategoryAutoStyler` and `CategoryColors`?

alonsogarciapablo:
I left some inline comments. The most important thing that needs to be addressed IMO is removing the need to know how CartoDB.js is rendering the map (i.e: all the references to `dataview._dataProvider`...).

src/widgets/auto-styler.js
```diff
@@ -0,0 +1,141 @@
+var cdb = require('cartodb.js');
```
alonsogarciapablo: This class deserves a unit test... There's a lot of "meat" inside which is untested.

src/widgets/category/category-colors.js
```diff
@@ -1,5 +1,8 @@
 var _ = require('underscore');
-var colors = ['#2CA095', '#E5811B', '#4A4DBA', '#AD2BAD', '#559030', '#E1C221']; // Demo colors
+var colorScales = [['#2CA095', '#E5811B', '#4A4DBA', '#AD2BAD', '#559030', 'rgb(166,206,227)', 'rgb(31,120,180)', 'rgb(178,223,138)', 'rgb(51,160,44)', 'rgb(251,154,153)', 'rgb(27,158,119)', 'rgb(217,95,2)', 'rgb(117,112,179)', 'rgb(231,41,138)', 'rgb(102,166,30)', 'rgb(230,171,2)', '#E1C221'],
```
javisantana: these color scales should be get from turbocartocss or the repository where all the new color scales are cc @andrewxhill @makella

fdansv: @javisantana turbocartocss doesn't support qualitative colouring, as far as I know

makella: @fdansv @javisantana @andrewxhill we'll have some new qualitative palettes ready in the next day. If you have any suggestions on the best way to provide those, please let me know. Right now we have the quantitative palettes (sequential not diverging _yet_) here: https://github.com/CartoDB/CartoColor/blob/master/cartocolor.js

src/widgets/category/category-widget-model.js
fdansv: ✅

src/widgets/category/category-widget-model.js
fdansv: ✅

src/widgets/category/category-widget-model.js
javisantana: it should be in widgetmodel since it's widget state (and needs to be serialized)

src/widgets/widgets-collection.js
```diff
@@ -15,12 +15,12 @@ module.exports = Backbone.Collection.extend({
     this.bind('change:order', _.debounce(this._onChangeOrder.bind(this), 10), this);
     // If a category model applies the category colors, rest should remove/disable
     // the category colors applied before.
-    this.bind('change:isColorsApplied', function (m, isColorCategorized) {
+    this.bind('change:isAutoStyle', function (m, isColorCategorized) {
```
alonsogarciapablo: I'm a bit confused about the `isAutoStyle` attribute... We're setting [the `autoStyle` attribute in the dataview](https://github.com/CartoDB/deep-insights.js/pull/304/files#diff-9174d99f35e6d799c1c69c256b44b3c3R59), but I don't see where (apart from the tests we are setting `isAutoStyle`)...

src/widgets/category/list/item/item-view.js
```diff
@@ -47,7 +47,8 @@ module.exports = cdb.core.View.extend({
 
   _initBinds: function () {
     this.model.bind('change', this.render, this);
-    this.widgetModel.bind('change:search change:isColorsApplied change:prefix change:suffix', this.render, this);
+    this.widgetModel.bind('change:search change:isAutoStyle change:prefix change:suffix', this.render, this);
+    this.dataviewModel.bind('change:autoStyle', this.render, this);
```
alonsogarciapablo: Do we need to listen to both the `change:isAutoStyle` event on the widgetModel and the `change:autoStyle` event on the dataviewModel?

src/widgets/histogram/histogram-title-view.js
```diff
@@ -38,7 +40,10 @@ module.exports = cdb.core.View.extend({
 
   _initBinds: function () {
     this.widgetModel.bind('change:title change:collapsed', this.render, this);
-    this.dataviewModel.bind('change:histogram_sizes', this.render, this);
+    this.dataviewModel.bind('change:autoStyle', function () {
+      this.render();
+      this.dataviewModel._onStyleChanged();
```
alonsogarciapablo: This should be handled by CartoDB.js internally IMO. If `autoStyle` has changed, it means that the `cartoCSS` has been changed, right? Can we "detect" this in CartoDB.js at the layer level and reload the map if needed?

spec/widgets/auto-style.spec.js
```diff
@@ -0,0 +1,30 @@
+var specHelper = require('../spec-helper');
```
alonsogarciapablo: Now that we have some nice subclasses of the different types of auto stylers, I think it makes sense to actually unit test each type and make sure that `.getStyle` generates the stuff right. That's where things can go wrong, so that's the method I would unit test. Here's a suggestion of how I'd structure a test for the category auto styler:

```
var CategoryAutoStyler = require('../../auto-style/category.js')

describe('src/widgets/auto-style/category', function () {
  beforeEach(function () {
    this.dataview = new Backbone.Model({
      allCategoryNames: 'a', 'b', 'c',
      column: 'something' 
    })

    this.layer = this.dataview.layer = jasmine.createSpyObj('layer', ['getGeometryType']);
    this.categoryAutoStyler = new CategoryAutoStyler({
      dataview: this.dataview
    });
  });

  describe('.getStyle', function () {
    it('should generate the right styles when layer has polygons', function () {
      this.layer.getGeometryType.and.returnValue('polygon');

      expect(this.categoryAutoStyler.getStyle()).toEqual('EXPECTED_STYLE');
    })

    it('should generate the right styles when layer has points', function () {
      this.layer.getGeometryType.and.returnValue('point');

      expect(this.categoryAutoStyler.getStyle()).toEqual('EXPECTED_STYLE');
    })

    it('should generate the right styles when layer has lines', function () {
      this.layer.getGeometryType.and.returnValue('line');

      expect(this.categoryAutoStyler.getStyle()).toEqual('EXPECTED_STYLE');
    })

    it('should generate the right styles when layer\'s geometry cannot be infered', function () {
      this.layer.getGeometryType.and.returnValue(undefined);

      expect(this.categoryAutoStyler.getStyle()).toEqual('EXPECTED_STYLE');
    })
  });
});
```

src/widgets/category/list/items-view.js
```diff
@@ -38,7 +38,7 @@ module.exports = cdb.core.View.extend({
 
   _initBinds: function () {
     this.widgetModel.bind('change:search', this.toggle, this);
-    this.widgetModel.bind('change:isColorsApplied', this.render, this);
+    this.widgetModel.bind('change:isAutoStyle', this.render, this);
```
alonsogarciapablo: `isAutoStyle`?

spec/widgets/widgets-collection.spec.js
```diff
@@ -12,27 +12,12 @@ describe('widgets/widgets-collection', function () {
...
     this.collection.reset([
-      {order: 0, isColorsApplied: false},
+      {order: 0, autoStyle: false},
       {order: 1},
-      {order: 2, isColorsApplied: true}
+      {order: 2, autoStyle: true}
     ]);
   });
 
-  it('should only allow applying colors in one widget at a time', function () {
```
alonsogarciapablo: Why are we getting rid of this test? Wasn't it testing [this](https://github.com/CartoDB/deep-insights.js/pull/304/files#diff-8165807a4462edc13831b72bc1328180L18)? IMO it's still important and replacing 'isColorApplied' by 'autoStyle' should make it pass, right?

fdansv: Because the premise the test is based on is not valid anymore. We do allow auto style to be affecting more than one widget at a time.

alonsogarciapablo: Ok, then what is [this](https://github.com/CartoDB/deep-insights.js/pull/304/files#diff-8165807a4462edc13831b72bc1328180L18) doing?

alonsogarciapablo:
👍  👍  👍

PR #314, opened 2016-04-26
added standalone index to have deepInsights scope inside cartodb one

### LINKED ISSUE #312: Prepare DI with cartodb.js publicly available
https://github.com/CartoDB/deep-insights.js/pull/311 removes cartodb.js API from DI.

[Browserify task relies on src/index.js](https://github.com/CartoDB/deep-insights.js/blob/332dc7434cc35c3a1a1cc650ec2182d4ce4f3390/grunt-tasks/browserify.js#L3) as entry point for generating the bundle but after #311 is merged it won't be available.

We need to create a new entry point that mixes cartodb.js and DI so the new final bundle is similar to the current one, where cdb.deepInsights exposes DI and cdb exposes cartodb.js in general.

javisantana:
fixes #312

cc @rochoa

### DIFF
--- a/grunt-tasks/browserify.js
+++ b/grunt-tasks/browserify.js
@@ -6,7 +6,7 @@ module.exports = {
       watch: '<%= config.doWatchify %>',
       browserifyOptions: {
         debug: true, // to generate source-maps
-        standalone: 'cartodb',
+        standalone: 'cartodb'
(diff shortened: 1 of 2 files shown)


rochoa:
🚢  🇮🇹

PR #325, opened 2016-04-28
Implements dashboard state set/get

fdansv:
@javisantana 

- Adds default states for all widgets (`collapsed` and `pinned`).
- Adds a getter and setter of state for widgets.
- The dashboard sweeps all the widgets' states and serialises them into a URL.
- The dashboard will initialise the widgets with the attributes specified when adding a `state` query parameter to the URL.


src/widgets-service.js
```diff
@@ -1,4 +1,5 @@
 var _ = require('underscore');
+var URI = require('urijs');
```
javisantana: I udnerstand is easy to use a npm module but I'd try to avoid to use libraries for things as easy as `location.hash.slice(1)`

src/widgets/widget-model.js
```diff
@@ -38,5 +43,20 @@ module.exports = cdb.core.Model.extend({
...
     this.stopListening();
+  },
+
+  setState: function (state) {
+    this.set(state);
+  },
+
+  getState: function () {
+    var state = {};
+    for (var key in this.defaultState) {
```
javisantana: why don't use `_.defaults`

src/widgets-service.js
```diff
@@ -150,6 +151,16 @@ WidgetsService.prototype.createTimeSeriesModel = function (attrs, layer) {
   return widgetModel;
 };
 
+WidgetsService.prototype.setWidgetsState = function (state) {
```
javisantana: the deserilization should be done outside here, this function should not know where it's serialized (how would you do if the state comes from an ajax request instead of the url?)

src/widgets/widget-model.js
javisantana: or better, why don't initialize the model with that state?

spec/api/dashboard.js
```diff
@@ -0,0 +1,18 @@
+var Dashboard = require('../../src/api/dashboard');
```
alonsogarciapablo: We should rename this file to "spec/api/dashboard.spec.js"

src/widgets-service.js
fdansv: I agree, but there are a couple more uses than just getting the query parameter, see `src/api/dashboard`

src/widgets/widget-model.js
fdansv: `_.default` does the job of filling the `undefined` blanks, but it doesn't discard values that are equal to the default (I'm adopting the policy of not adding unchanged attributes to the state)

src/widgets-service.js
fdansv: I agree that the URL logic would probably better be elsewhere, but it only takes the state from it if there isn't a `state` argument passed to the function. I suppose an ajax request would use `setWidgetsState(state_object_from_ajax_req)`?

fdansv: also I'm guessing that the ajax request would use `dashboard.setState()`, not the function in `WidgetsService`

javisantana: there should be a state object, yes

src/widgets-service.js
javisantana: dashboard should not know about urls anyways, it's a embed/html responsability (maybe using a helper). Anyways, I would use the simplest solution to avoid npm module bloating (and complains about javascript file size :blink: :blink:)

fdansv:
(merged after offline ok from @javisantana)

PR #354, opened 2016-05-17
Default state to normalized.

### LINKED ISSUE #352: Switcher on normalize should be activated by default.
Go to any dashboard, click on any histogram widget's three dots and see:

Even having the graph normalized by the total, the switcher looks the activated.
Right now it works in the opposite way that it should be working.

@nobuti no hay huevos.

nobuti:
This PR fixes #352.

Setting explicitly the default state to normalized in the widget model makes the deal. The chart is already rendered initially as normalized (in https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/histogram/chart.js#L526 `this._originalYScale` is undefined the first time), so the UI needs the normalized state to be rendered in sync.

cc @javisantana @xavijam


javisantana:
cc @nobuti review changes please

PR #360, opened 2016-05-23
Adds setters and getters to the map state

fdansv:
Now the map position (center coordinates and zoom level) are returned as part of the state object, and as a shareable url.

@javisantana


javisantana:
:es:

PR #361, opened 2016-05-24
Export reference to the dashboard view.

nobuti:
In https://github.com/CartoDB/cartodb/issues/7445, we need a public reference to the view, in order to change the background color of the dashboard view when we are editing.

cc @xavijam


xavijam:
👍

PR #362, opened 2016-05-24
Category & Histogram state set/get

fdansv:
* [ ] Merge https://github.com/CartoDB/cartodb.js/pull/1284

@javisantana

### DIFF
--- a/spec/widgets/category/item-view.spec.js
+++ b/spec/widgets/category/item-view.spec.js
@@ -26,4 +26,10 @@ describe('widgets/category/item-view', function () {
     this.view.render();
     expect(this.view.$('.CDB-Widget-progressState').hasClass('is-accepted')).toBeTruthy();
   });
+
+  it('should update the widget\'s state accordingly when selected', function () {
+    this.dataviewModel.filter.accept('USA');
+    this.view.render();
+    expect(this.widgetModel.getState().acceptedCategories[0]).toEqual('USA');
+  })
 });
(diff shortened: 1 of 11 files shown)


javisantana:
My main concern here is the state serialization should be done (or at least be able to do it) outside of deep-insights

In other words, the responsable of serialize state from/to url should be someone outside DI.js, for example the embed.

The reason is clear, we will want to serialize states from urls but from database as well and the current designs avoids it.

There is a simple change that improvides this a lot

```js
// var stateFromURL = URLHelper.getStateFromCurrentURL();
var stateFromURL = opts.state || URLHelper.getStateFromCurrentURL();
```
not perfect but cover that use case

There is another important thing missing, versioning. When we update widgets there will be more things, having a version will allow us to check if the state is valid. This is minor but take that into account

fdansv:
> My main concern here is the state serialization should be done (or at least be able to do it) outside of deep-insights

Yeah that makes sense, although the scope for this feature was *dashboard state serialization*, so if we're thinking in terms of dashboards it made sense to apply it to DI.js

> There is a simple change that improvides this a lot
Ha, somewhere a few commits ago in this PR I had that precisely. Bear in mind, though, that we also have `dashboard.setState()`, which could be called from the embed, right?

javisantana:
> Ha, somewhere a few commits ago in this PR I had that precisely. Bear in mind, though, that we also have dashboard.setState(), which could be called from the embed, right?

Yep, that's fine but is it possible to initialize the dashboard with a state?

fdansv:
OK I see, I just added that

PR #367, opened 2016-06-06
Only encodes state string once

fdansv:
State strings in URLs seem to be URLencoded twice, because the library that we use to parse URLs already does the encoding internally. I'm just removing our call here.
@xavijam


xavijam:
:+1:

PR #374, opened 2016-06-17
Uses layerX instead of offsetX to calculate clicked bar index

### LINKED ISSUE #373: When you click over a bin in a histogram, it goes to the first one.
histogram-click-bug

fdansv:
Using offsetX was causing the `_getBarIndex` to return always 0.

Fixes #373 

CR @javierarce 🙌


javierarce:
👍

PR #376, opened 2016-06-22
Keeps marker widths on autostyle 

fdansv:
Notice that this applies to turbo-carto and static marker size rules, like the following:
```
marker-width: ramp([col],10, 30, 'jenks');
```
or
```
marker-width: 10;
```

### DIFF
--- a/src/widgets/auto-style/auto-styler.js
+++ b/src/widgets/auto-style/auto-styler.js
@@ -16,10 +16,10 @@ var AutoStyler = cdb.core.Model.extend({
   getPreservedWidth: function () {
     var startingStyle = this.layer.get && this.layer.get('cartocss');
     if (startingStyle) {
-      var originalWidth = startingStyle.match(/marker-width:.*;/g);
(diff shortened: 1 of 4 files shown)


src/widgets/auto-style/auto-styler.js
```diff
@@ -11,6 +11,21 @@ var AutoStyler = cdb.core.Model.extend({
...
+      var originalWidth = startingStyle.match(/marker-width:.*?;\s/g);
+      if (originalWidth) {
+        if (originalWidth.length > 1) {
+          return null;
+        } else {
+          originalWidth = originalWidth[0].replace('marker-width:', '').replace(';', '');
+        }
+      }
+    }
+    return originalWidth;
```
javierarce: if `startingStyle` is not defined, `originalWidth` won't be declared.

fdansv: Ouch yeah, missed that in the mini refactor

javierarce:
Except for my comment, the rest is 👍

PR #382, opened 2016-06-23
Bars coloring according to histogram autostyling

fdansv:


xavijam:
Taking care of it due to the fact that I'll assume all auto-style stuff, 💋  @fdansv.

fdansv:
@xavijam it's a tricky one ;) . As it is right now the bars are being coloured according to the palette passed by the autostyling, but it is not taking into account the quantification method, so colours in the histogram don't always match the values in the map. 

This is easy to circumvent with client side rendering since quantification is easily accessible, but I wasn't sure how to do it with the main case, which is server-side rendering.

xavijam:
Agree @fdansv, it is not an easy task. We are gonna just "represent" the colors in the histogram bars although any of them will not be 100% accurate. 🤗

PR #388, opened 2016-06-28
Default states fixes

fdansv:
* Histogram bars are not passed in the state if they haven't been changed
* Map state will not be reported if the position hasn't changed from 1% of its default position.

### DIFF
--- a/src/api/dashboard.js
+++ b/src/api/dashboard.js
@@ -49,12 +49,13 @@ Dashboard.prototype = {
     var upperLon = initialState.center[1] * 0.01;
     var inside = (Math.abs(initialState.center[0] - currentCenter[0]) < upperLat) && (Math.abs(initialState.center[1] - currentCenter[1]) < upperLon);
     if (inside && _.isEqual(currentZoom, initialState.zoom)) {
+      return {};
+    } else {
       return {
         center: currentCenter,
         zoom: currentZoom
       };
     }
-    else return {};
   },
 
   setState: function (state) {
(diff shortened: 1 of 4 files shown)


fdansv:
@alonsogarciapablo mind taking a quick look? ;)

src/api/dashboard.js
```diff
@@ -45,11 +45,17 @@ Dashboard.prototype = {
     var initialState = this._dashboard.dashboardView.getInitialMapState();
     var currentCenter = this._dashboard.vis.map.get('center');
     var currentZoom = this._dashboard.vis.map.getZoom();
-    if (_.isEqual(currentCenter, initialState.center) && _.isEqual(currentZoom, initialState.zoom)) return {};
-    return {
-      center: currentCenter,
-      zoom: currentZoom
-    };
+    var upperLat = initialState.center[0] * 0.01;
```
alonsogarciapablo: I'd be easier for me to understand what this change is about if there was a test for it...

alonsogarciapablo: What is `0.01`? I read the PR description and I know that is a 1% of something... but my (or even your) future self might not remember what this is... Could we declare a "constant" and give a name?

alonsogarciapablo: Also, we could make this a bit clearer/easier to understand by extracting a method:

```
if (this._hasStateChanged(...)) {
  return {
    center: currentCenter,
    zoom: currentZoom
  };
} else {
  return {};
}
```

alonsogarciapablo:
It makes sense. I left some comments/suggestions to make it a bit clearer and I wish there were some tests (specially for the part that implements `Map state will not be reported if the position hasn't changed from 1% of its default position`).

PR #394, opened 2016-06-28
Inverting the autostyle workflow so styles aren't rewritten, but modified.

fdansv:
This gets rid of all the problems we've had related to cartocss changing dramatically when hitting autostyle. Now, instead of using predefined templates, we just modify some lines in the cartocss and pass it to the tiler as is.

@javisantana @makella 

sickauto


javisantana:
there is a better way to override styles:

```less
#layer {
  marker-width: 10;
  marker-fill: red;

}

#layer {
  marker-width: 2;
  marker-fill: blue;
}
```

in this case mapnik will render marker-width: 2 in blue.

Let's merge this as-is and test how it works as soon as possible. if that works we can think in a better way to do this

PR #401, opened 2016-07-03
Fixes bars not showing in responsive, small histogram

### LINKED ISSUE #400: When turning to a small, responsive histogram, the bars won't draw
reac

It appears this is because of two factors:
* The chart is not updating it's Y scale each time the histogram is reset.
* The minimum width is set to zero right now, which causes bars not to show.

fdansv:
Fixes #400 , related to https://github.com/CartoDB/cartodb/issues/8320

@javisantana


javierarce:
👍

PR #402, opened 2016-07-04
fix embed ui

piensaenpixel:

image

cc/ @saleiva


src/dashboard-view.js
```diff
@@ -48,7 +49,7 @@ module.exports = cdb.core.View.extend({
     });
     this.addView(view);
     this.$el.append(view.render().el);
-
+    $('body').addClass('CDB-Embed-container');
```
xavijam: Add a comment about why we add this thing, please...

xavijam:
:+1:

PR #403, opened 2016-07-05
createDashboard handles errors diferently

alonsogarciapablo:
`createDashboard` was throwing an error, which was preventing the UI from being rendered.

Changes:
  - createDashboard doesn't throw error and forwards error to the callback instead.
  - Usage of `vis.instantiateMap` now gets an object with `success` and `error` callbacks instead of using `vis.done` and `vis.error` (which were not working). See: https://github.com/CartoDB/cartodb.js/pull/1334

@xavijam sounds good?


xavijam:
Music to my ears.

PR #404, opened 2016-07-05
Remove url state fetch

gfiorav:
Dashboard now relies only on whatever state is provided through options.

The decision to take the state from the URL or elsewhere is now taken by the Cartodb UI. The dashboard does retain the capacity of generating urls though (`getDashboardURL`) and make them copiable and sharable :).

This can be merged when https://github.com/CartoDB/cartodb/pull/8507 is merged.

Please, @fdansv have a look!

cc/ @javisantana

### DIFF
--- a/src/api/create-dashboard.js
+++ b/src/api/create-dashboard.js
@@ -47,11 +47,12 @@ var createDashboard = function (selector, vizJSON, opts, callback) {
     widgets: widgets,
     model: model
   });
-  var stateFromURL = opts.state || URLHelper.getStateFromCurrentURL();
-  if (!_.isEmpty(stateFromURL.map)) {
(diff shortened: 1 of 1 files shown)


fdansv:
The only thing here is that standalone deep-insights won't be collecting states from the URL, but I guess that's ok?

Other that that, I've no objections.

🚢 🇮🇹

gfiorav:
Thanks for taking a look @fdansv. Certainly, that is the drawback. Actually, looking at it now, we can also not do this, and it will prefer whatever is in the opts.state over the URL, which would still work in both cases.

What's your take, @javisantana ?

gfiorav:
Just spoke with @javisantana, let's leave this like this since we don't need it!

Thanks!

PR #406, opened 2016-07-05
Rounds default numeric values

fdansv:
Sometimes the values returned by the tiler differ from the ones in the dataview by decimal positions. This isn't really solving the problem, but it is a hotfix to avoid setting wrong values in the state.

Fixes https://github.com/CartoDB/cartodb/issues/8460

@javisantana

### DIFF
--- a/src/widgets/histogram/histogram-widget-model.js
+++ b/src/widgets/histogram/histogram-widget-model.js
@@ -10,13 +10,7 @@ module.exports = WidgetModel.extend({
   defaultState: _.extend(
     {
       autoStyle: false,
-      normalized: false,
-      min: function () {
-        return this.dataviewModel.get('start');
-      },
-      max: function () {
-        return this.dataviewModel.get('end');
-      }
+      normalized: false
     },
     WidgetModel.prototype.defaultState
   ),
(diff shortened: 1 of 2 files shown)


javisantana:
why not using an epsilon?

Rounding a number like 0.0001 is kind of risky

javisantana:
- I don't think we can make this generic, it should be specific for histogram
- we can't just compare with 0.1, it should be some % of the histogram range because for ranges less than 0.1 we are breaking the serialization state. If we don't want to do that far we should use a Number.EPSILON or something like that

fdansv:
Updated it using Number.EPSILON

src/widgets/histogram/histogram-widget-model.js
```diff
@@ -53,6 +47,27 @@ module.exports = WidgetModel.extend({
...
+  },
+
+  getState: function () {
+    var state = WidgetModel.prototype.getState.call(this);
+    var start = this.dataviewModel.get('start');
+    var end = this.dataviewModel.get('end');
+    var min = this.get('min');
+    var max = this.get('max');
+    var checkRoughEqual = function (a, b) {
+      if (_.isNumber(a) && _.isNumber(b) && Math.abs(start - min) > Number.EPSILON) {
```
javisantana: I though the problem was checking the actual state with the serialized one not the histogram range, shouldn't be this Math.abs(a - b) ?

fdansv: yes, omg, my brain wasn't working at that point. Correcting now.

PR #409, opened 2016-07-07
fix problem with scroll

piensaenpixel:
Fix the problem with the scroll  and resize the window, and minor ui adjust

scroll

cc/@saleiva

### DIFF
--- a/src/dashboard-sidebar-view.js
+++ b/src/dashboard-sidebar-view.js
@@ -68,6 +68,8 @@ module.exports = cdb.core.View.extend({
     this._renderShadows();
     this._bindScroll();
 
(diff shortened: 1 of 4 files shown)


src/dashboard-sidebar-view.js
```diff
@@ -68,6 +68,8 @@ module.exports = cdb.core.View.extend({
     this._renderShadows();
     this._bindScroll();
 
+    $(window).on('resize', this._onResize.bind(this));
```
xavijam:
```js
$(window).on('resize', _.debounce(this._onResize.bind(this), 1000));
```
What do you think @nobuti ?

xavijam:
:+1:

PR #412, opened 2016-07-07
Adding urls without params + target blank for user info

xavijam:
Twitter can't share an url super long (probably due to the state url feature), so let's share in Facebook and Twitter urls without params.

CR: @javisantana 
cc @piensaenpixel


xavijam:
Ping!

piensaenpixel:
vamos!

PR #421, opened 2016-07-13
Flexible precision for small and negative values

### LINKED ISSUE #413: Too much rounding on histogram when values have little differences
screen shot 2016-07-11 at 17 49 50

this is for values between 0.991 and 1.

saleiva:
close #413

### DIFF
--- a/spec/formatter.spec.js
+++ b/spec/formatter.spec.js
@@ -3,10 +3,15 @@ var formatter = require('../src/formatter');
 describe('formatter', function () {
   it('should format numbers', function () {
     expect(formatter.formatNumber(0)).toBe('0');
+    expect(formatter.formatNumber(0.71)).toBe('0.710');
+    expect(formatter.formatNumber(-0.71)).toBe('-0.710');
     expect(formatter.formatNumber(5)).toBe('5');
     expect(formatter.formatNumber(5.0)).toBe('5');
     expect(formatter.formatNumber(5.00)).toBe('5');
-    expect(formatter.formatNumber(186.7)).toBe('186.70');
+    expect(formatter.formatNumber(5.71)).toBe('5.71');
+    expect(formatter.formatNumber(-5.71)).toBe('-5.71');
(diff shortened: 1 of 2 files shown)


saleiva:
CR @javisantana 
I cannot make the failing test to pass - fucking js

piensaenpixel:
Lint is broken, i can try tomorrow

saleiva:
retest please

saleiva:
CR @javisantana

spec/formatter.spec.js
```diff
@@ -2,11 +2,16 @@ var formatter = require('../src/formatter');
 
 describe('formatter', function () {
   it('should format numbers', function () {
-    expect(formatter.formatNumber(0)).toBe('0');
+    expect(formatter.formatNumber(0)).toBe(0);
+    expect(formatter.formatNumber(0.71)).toBe('0.710');
```
javisantana: why 0.710 and not 0.71 ?

src/formatter.js
```diff
@@ -4,29 +4,36 @@ var d3 = require('d3');
...
 
-  value = format(value) + (unit ? ' ' + unit : '');
+  if (abs_v > 100) {
+    p = 0;
+  } else if (abs_v > 10) {
+    p = 1;
+  } else if (abs_v > 1) {
+    p = 2;
+  } else if (abs_v > 0) {
+    p = 3;
```
javisantana: if abs_v = 0.0001 then p = 3 so 0.0001.toFixed(3) is 0.000 which does not make any sense.

I'd use a logarigthm here (when under 1) to see how many of them I want, like

Math.ceil(Math.abs(Math.log(0.0001)/Math.log(10))) -> 4



toFixed( 4 + 2)

saleiva: That would likely not work fine with numbers like 0.3141 where the result would be 1

saleiva:
CR @javisantana

javisantana:
it lacks of a test for small numbers :/

saleiva:
added those @javisantana

PR #425, opened 2016-07-15
Only show points on search pagination if pages > 1

### LINKED ISSUE #408: Auto-style fails when having # on category names
screen shot 2016-07-06 at 14 20 32

Within the builder, styling by category on the same column works correctly

saleiva:
close ##398


saleiva:
CR @javisantana

javisantana:
:es:

PR #432, opened 2016-07-20
Removes unnecessary function and avoids unnecessary calls to updateStats when changing the range on the histogram

saleiva:
CR @javierarce @xavijam 

This was motivated by some debugging I did, where I noticed several repeated calls

### DIFF
--- a/src/widgets/histogram/content-view.js
+++ b/src/widgets/histogram/content-view.js
@@ -166,7 +166,6 @@ module.exports = cdb.core.View.extend({
     this.$('.js-content').append(this.histogramChartView.el);
     this.addView(this.histogramChartView);
 
-    this.histogramChartView.bind('range_updated', this._onRangeUpdated, this);
     this.histogramChartView.bind('on_brush_end', this._onBrushEnd, this);
     this.histogramChartView.bind('hover', this._onValueHover, this);
     this.histogramChartView.render().show();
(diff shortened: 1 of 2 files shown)


src/widgets/histogram/content-view.js
```diff
@@ -255,8 +254,7 @@ module.exports = cdb.core.View.extend({
 
   _onBrushEnd: function (loBarIndex, hiBarIndex) {
     var data = this._dataviewModel.getData();
-
-    if (!data || !data.length) {
+    if ((!data || !data.length) || (this.model.get('lo_index') == loBarIndex && this.model.get('hi_index') == hiBarIndex)) {
```
saleiva: Avoid executing the function if the new range is the same as the one set previously

xavijam:
Tests are not passing, take a look please. Also, I don't know how exactly histogram with brush works but I have a question. Have you tried this with an example using the state of the url?

saleiva:
Yes yes, reviewing the tests

saleiva:
In local they pass

xavijam:
it is a lint error, please check travis details.

xavijam:
:es:

saleiva:
@javierarce @xavijam doesn't trust me until you review this

javierarce:
Haha, sorry, code looks good to me 👍

PR #433, opened 2016-07-20
fixing problem with palette names

saleiva:


saleiva:
@makella I've just discovered we were using old names. It's very important that we don't change those or backwards compatibility will fail.

CR @javisantana

javisantana:
using where?

javisantana:
but we'd need to update cartocolors or something, right?

makella:
yeah totally i knew this would break things but we didn't have a chance to update deep insights yet. i'm sorry!

makella:
@saleiva one other thing the palettes deep insights was using were from April! So there were updates that needed to happen.

saleiva:
ok, can you comment on those updates on tickets?

makella:
In the future? Or would you like me to do that here now?

I opened the question back in May in Carto Color repo... probably the wrong place. 

Sorry!!

https://github.com/CartoDB/CartoColor/issues/5

Issue #49, opened 2015-12-30
Make "pin widget" action work


xavijam:
![ebc183aa-ad53-11e5-8719-da80764cf713](https://cloud.githubusercontent.com/assets/132146/12049789/7a332b40-aeec-11e5-97fb-b624a0414345.png)

Delete option should be only available in editor.

noguerol:
Still not working cc @saleiva

PR #437, opened 2016-07-22
Removes pinned option from widgets until implementing it

saleiva:


saleiva:
CR @javierarce

saleiva:
close #49

xavijam:
:+1:

PR #440, opened 2016-08-08
Fixing some of the problems related with state and histogram

### LINKED ISSUE #416: Histogram selection is not being properly restored from URL
# STR:

1. Go into the embed of a map with a histogram widget
2. Apply a selection on the histogram
3. Refresh

# Expected result:

The selection is reapplied.

# Actual result:

No selection is visible in the histogram (it's appearance is exactly the same as before applying any selection) but the map visualization *is* indeed filtered.

xavijam:
Several problems identified for the histogram:

- [ ] `start` and `end` were needed in order to set the proper bounds.
- [ ] If there were any state applied, it provoked problems trying to set a new state (due to the change binding) without initializing the first state (workaround trying to identify if that was the only problem).
- [ ] `start` and `end` should not be requested after the map instantiation, if not filter could change the returned data (@rochoa).
- [ ] After a second refresh with histrogram state changes, the histogram values disappear.

Trying to fix #416.

cc @gfiorav

### DIFF
--- a/src/api/dashboard.js
+++ b/src/api/dashboard.js
@@ -65,8 +65,25 @@ Dashboard.prototype = {
   },
 
   onStateChanged: function (callback) {
-    this._dashboard.widgets._widgetsCollection.bind('change', function () {
-      callback(this.getState(), this.getDashboardURL());
+    var firstLoad = true;
+    var widgets = this._dashboard.widgets._widgetsCollection;
+
+    widgets.bind('change', function () {
+      var triggerCallback = true;
+
+      if (firstLoad) {
+        var widgetsNotLoaded = _.filter(widgets.dataviewModel, function (mdl) {
+          return mdl._pending === true;
+        });
+        console.log(arguments[0], widgetsNotLoaded);
+        if (widgetsNotLoaded.length !== 0) {
+          triggerCallback = false;
+        } else {
+          firstLoad = false;
+        }
+      }
+
+      triggerCallback && callback(this.getState(), this.getDashboardURL());
     }, this);
     this._dashboard.vis.map.bind('change', function () {
       callback(this.getState(), this.getDashboardURL());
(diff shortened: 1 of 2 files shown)


src/api/dashboard.js
```diff
@@ -65,8 +65,25 @@ Dashboard.prototype = {
...
+        });
+        console.log(arguments[0], widgetsNotLoaded);
+        if (widgetsNotLoaded.length !== 0) {
+          triggerCallback = false;
+        } else {
+          firstLoad = false;
+        }
+      }
+
+      triggerCallback && callback(this.getState(), this.getDashboardURL());
```
viddo: I think this approach is error-prone, using private variables and leaving this event listener around since it will modify those temporary variables even after the initial events, which could hamper the expected behavior. Let's not go down this path.

I think part of the problem is related to how this _state_ is handled; The initial state should be given as _attributes_, it doesn't really make sense to set "initial state" right after the model was created. I'd remove the setState method and solve the initial-state problem to avoid events that should not event fire initially.

src/widgets/histogram/histogram-widget-model.js
```diff
@@ -52,6 +52,19 @@ module.exports = WidgetModel.extend({
     this.set('autoStyle', false);
   },
 
+  setState: function (state) {
+    if (state && state.start && state.end) {
```
viddo: From the PR desc:
> start and end were needed in order to set the proper bounds.
> …
> start and end should not be requested after the map instantiation, if not filter could change the returned data (@rochoa).
> …
> After a second refresh with histrogram state changes, the histogram values disappear.

Not familiar enough with the histogram model and what it does, I think @javierarce knows that part better. 
That said, from what I can see in the implementation I think part of the problem is related to my [other comment](https://github.com/CartoDB/deep-insights.js/pull/440/files#r74257959), assuming that part would be fixed, the histogram model should setup the initial state (request full start/end range etc.) _regardless_ of if the initial attrs contains a custom range or not. If the attrs contains a custom range it should of course apply that as well.

xavijam:
cc @nobuti, I feel it could help in order to solve correctly the state-url feature.

PR #443, opened 2016-08-11
Disable widget state

gfiorav:
This disables widget state until [the issues with it](https://github.com/CartoDB/deep-insights.js/issues/416) are fixed. Until then, only map related state is shared (i.e, viewport).

Please CR @xavijam


xavijam:
👍

PR #455, opened 2016-10-05
Using proper dependencies for browserify-resolutions

xavijam:
It fixes the problem we have right now with the bundle generation.

cc @nobuti


nobuti:
🎉

xavijam:
Default tests passing, merging!

PR #461, opened 2016-10-10
Complete cleaning for repetitives symbolizers fix #457

### LINKED ISSUE #457: Autostyling working in Builder but not on embed
@rochoa 

I made a map:
https://team.carto.com/u/mamataakella/builder/92096530-ccf0-4a97-8a29-7b0c2fe89061/embed

If you click on any of the Autostyling options in the widgets, you will notice unexpected behavior. Some polygons are getting colored with the autostyle colors and others are not.

screen shot 2016-10-05 at 9 53 46 am

Now, when I do the Autostyling inside of Builder, it works as expected:

screen shot 2016-10-05 at 9 54 55 am

So, I'm not sure what is happening between the Builder map and the embed.

Any help is appreciated!

Table:
https://team.carto.com/u/mamataakella/dataset/final_election_subquery_copy

donflopez:
We are replacing the attrs in other way. This fix #457 issue.

- Mark the attr to change

- Remove all of them

- Remove empty layers or conditions

- Change the mark with the new value of the attr

I've put the functional code just above, but it is not reusable and this is not a good place for it. We have to move it to a better place aligned with the project architecture to allow reusability and a better understanding.

This is a temporal solution, waiting for a better one using cartocss parser.

NOTE: I have to add tests to this PR, don't merge! 🙂 

cc @javisantana & @alonsogarciapablo

### DIFF
--- a/spec/widgets/auto-style/style-utils.spec.js
+++ b/spec/widgets/auto-style/style-utils.spec.js
@@ -1,0 +1,35 @@
+var StyleUtils = require('../../../src/widgets/auto-style/style-utils.js');
+
+describe('src/widgets/auto-style/style-utils', function () {
+  describe('.changeStyle', function () {
+    it('should change style attribute', function () {
+      var cartocss = '#layer {marker-line-color: white; marker-fill: #adadad;}';
+      var attr = 'marker-fill';
+      var newStyle = ' marker-fill: blue;';
+
+      var expected = '#layer {marker-line-color: white; marker-fill: blue;}';
+
+      expect(StyleUtils.changeStyle(cartocss, attr, newStyle)).toBe(expected);
+    });
+
+    it('should change style and remove duplicated', function () {
+      var cartocss = '#layer {marker-line-color: white; marker-fill: #adadad; marker-fill: #fafafa;}';
+      var attr = 'marker-fill';
+      var newStyle = ' marker-fill: blue;';
+
+      var expected = '#layer {marker-line-color: white; marker-fill: blue;}';
+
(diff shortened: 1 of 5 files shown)


donflopez:
Tests added!

spec/widgets/auto-style/style-utils.spec.js
```diff
@@ -0,0 +1,35 @@
...
+      var attr = 'marker-fill';
+      var newStyle = ' marker-fill: blue;';
+
+      var expected = '#layer {marker-line-color: white; marker-fill: blue;}';
+
+      expect(StyleUtils.changeStyle(cartocss, attr, newStyle)).toBe(expected);
+    });
+
+    it('should change style, remove duplicated and remove empty braces', function () {
+      var cartocss = '#layer {marker-line-color: white; marker-fill: #adadad; [me>gasol] { marker-fill: #fafafa; }}';
```
alonsogarciapablo: Would:
`var cartocss = '#layer {marker-line-color: white;  [me>gasol] { marker-fill: #fafafa; }; marker-fill: #adadad; }'` work too?

What about:

`'#layer {marker-line-color: white; marker-fill: #adadad; [me>gasol] { marker-fill: #fafafa; marker-width: 200}}'` ? (we could add a test to ensure layers that are not empty are not removed)

alonsogarciapablo:
About the approach... have you considered parsing the CartoCSS with http://github.com/CartoDB/carto instead of using regular expressions? I'm asking because that's probably what @javisantana was referring to in https://github.com/CartoDB/deep-insights.js/issues/457.

donflopez:
Yes, but we saw that we can't get cartocss from carto object, so we did this implementation until the carto -> cartocss logic will done.

PR #469, opened 2016-10-18
Get global distribution type to fix #468

### LINKED ISSUE #468: Widget auto-style color ramp should be always calculated with the total data and not with the bounding box
Sometimes when you are looking at all the data it applies a color ramp.
And then, when zooming in and doing so again, it applies a different color ramp.

Which is kind of confusing. The color ramp should be always calculated with all the data.

#SMOKE_WD15

cc/ @javisantana

donflopez:
This PR needs from this [other one.](https://github.com/CartoDB/cartodb.js/pull/1450)

It use global data to generate ramp color.

### DIFF
--- a/src/widgets/auto-style/histogram.js
+++ b/src/widgets/auto-style/histogram.js
@@ -12,7 +12,9 @@ var HistogramAutoStyler = AutoStyler.extend({
   },
 
   getColorLine: function (sym) {
-    var shape = this.dataviewModel.getDistributionType();
+    var shape = this.dataviewModel.getDistributionType(
+      this.dataviewModel.getUnfilteredDataModel().get('data')
+    );
     var scales = HistogramAutoStyler.SCALES_MAP[sym][shape];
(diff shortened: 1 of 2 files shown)


src/widgets/auto-style/histogram.js
```diff
@@ -12,7 +12,9 @@ var HistogramAutoStyler = AutoStyler.extend({
   },
 
   getColorLine: function (sym) {
-    var shape = this.dataviewModel.getDistributionType();
+    var shape = this.dataviewModel.getDistributionType(
+      this.dataviewModel.getUnfilteredDataModel().get('data')
```
javisantana: one comment, I don't know if unfiltered data model will have the whole data histogram when the dashboard stats with a filter (or other widget has a filter). I think it worths checking

javisantana:
:es:

src/widgets/auto-style/histogram.js
donflopez: Let me see! Thanks!

PR #473, opened 2016-10-24
State url

### LINKED ISSUE #445: State URL problems
- [x] We need to make a map instantiation _without filters_, in order to get the full data range. If there is no filters applied, go ahead and request all widgets data. If there is any filter, make a second map instantiation with the filters, and then request all the widgets data.
- [x] Don't set state URL when the map is initialized.
- [x] Don't listen widgets change until all widgets and map have been started (it triggers URL changes when state change).
- [x] Don't store center and zoom, better to store bounding box, `ne` and `sw` (https://github.com/CartoDB/deep-insights.js/issues/444).

nobuti:
This PR fixes #445 in DI's side. It's depends on https://github.com/CartoDB/cartodb.js/pull/1428.

cc @xavijam @donflopez

### DIFF
--- a/src/api/dashboard.js
+++ b/src/api/dashboard.js
@@ -36,10 +36,7 @@ Dashboard.prototype = {
     var state = {};
     var mapState = this.getMapState(); // TODO
     if (!_.isEmpty(mapState)) state.map = mapState;
-    /*
-     * TODO: Disabled widget states until issues are fixed.
-     * See https://github.com/CartoDB/deep-insights.js/issues/416
-     */
+
     var widgetsState = this._dashboard.widgets._widgetsCollection.getStates();
     if (!_.isEmpty(widgetsState)) state.widgets = widgetsState;
     return state;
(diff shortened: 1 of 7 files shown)


nobuti:
Ready for CR 🙏

src/api/dashboard.js
```diff
@@ -36,46 +36,36 @@ Dashboard.prototype = {
     var state = {};
     var mapState = this.getMapState(); // TODO
     if (!_.isEmpty(mapState)) state.map = mapState;
-    /*
-     * TODO: Disabled widget states until issues are fixed.
-     * See https://github.com/CartoDB/deep-insights.js/issues/416
-     */
-    // var widgetsState = this._dashboard.widgets._widgetsCollection.getStates();
-    // if (!_.isEmpty(widgetsState)) state.widgets = widgetsState;
+
```
xavijam: And cc @gfiorav, enabling it again! 💃  yay!

PR #475, opened 2016-10-25
fix #465

### LINKED ISSUE #465: Toggle is broken on the Histogram widget
The "Toggle widget" option, once activated, doesn't collapse the widget as expected. Actually, nothing happens. 

Also, on Safari, the toggle itself of that option, is misaligned. 

Already gave the deets to @piensaenpixel and I'm assigning this to him.

piensaenpixel:
@nobuti could you take a look?

image

Fix - https://github.com/CartoDB/deep-insights.js/issues/465

### DIFF
--- a/src/widgets/histogram/content-view.js
+++ b/src/widgets/histogram/content-view.js
@@ -49,6 +49,7 @@ module.exports = cdb.core.View.extend({
 
     this._renderMiniChart();
     this._renderMainChart();
+    this._renderAllValues();
   },
 
   _initTitleView: function () {
@@ -112,6 +113,8 @@ module.exports = cdb.core.View.extend({
 
   render: function () {
     this.clearSubViews();
+    this._unbinds();
+
     var data = this._dataviewModel.getData();
     var originalData = this._originalData.getData();
     var isDataEmpty = !_.size(data) && !_.size(originalData);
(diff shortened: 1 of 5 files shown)


nobuti:
collapse

nobuti:
collapse

@xavijam take a look and ping me if you find anything odd, please, I know you have been working on the histogram widget.

src/widgets/histogram/content-view.js
```diff
@@ -49,6 +49,7 @@ module.exports = cdb.core.View.extend({
 
     this._renderMiniChart();
     this._renderMainChart();
+    this._renderAllValues();
```
xavijam: Why do we need to do that?

xavijam: I mean, we used to follow the "show_stats" property in order to render the stats or not.

src/widgets/histogram/content-view.js
```diff
@@ -213,6 +217,18 @@ module.exports = cdb.core.View.extend({
     this.model.bind('change:avg', this._onChangeAvg, this);
   },
 
+  _unbinds: function () {
+    this._dataviewModel.off('change:bins', this._onChangeBins, this);
```
xavijam: I guess it is important when the histogram is rerendered by any option right?

src/widgets/histogram/content-view.js
nobuti: But the thing is if collapsed, we don't render the stats. If we uncollapse it, we render it back. But the values are set only on model change. We need to repaint them.

src/widgets/histogram/content-view.js
nobuti: Because we are calling `render` when collapsing, this function unbinds events in order to not duplicate bindings. Below `_setupBindings` is called again.

xavijam: :es:

xavijam:
:es:

PR #478, opened 2016-10-26
Histogram state

### LINKED ISSUE #445: State URL problems

nobuti:
This PR fixes #445.

CR @xavijam


nobuti:
On it's way to staging...

nobuti:
In ded05

PR #481, opened 2016-10-31
Fix 0 of null error in staging

donflopez:


src/widgets/histogram/chart.js
```diff
@@ -1002,11 +1002,22 @@ module.exports = cdb.core.View.extend({
...
+      sheet = sheets[i];
+      if (sheet.cssRules) {
+        rules = sheet.cssRules;
+        break;
+      }
+    }
 
-    if (document.styleSheets[0].cssRules[0].selectorText !== '.CDB-Chart-bar.is-highlighted') {
-      document.styleSheets[0].insertRule('.CDB-Chart-bar.is-highlighted { fill: ' + color + ' !important; }', 0);
+    if (rules[0].selectorText !== '.CDB-Chart-bar.is-highlighted') {
```
nobuti: what if `rules` doesn't exist?

donflopez: Yeah @nobuti! I forgot that check :)

Thank you!!

PR #483, opened 2016-10-31
Rollback

donflopez:


src/widgets/time-series/torque-histogram-view.js
```diff
@@ -55,7 +55,7 @@ module.exports = cdb.core.View.extend({
       animationBarDelay: function (d, i) {
         return (i * 3);
       },
-      chartBarColor: this._parent.model.getWidgetColor() || '#F2CC8F',
```
xavijam: When we merge it again to master, IMHO we should not use this notation, `_parent.model` is a smell.

PR #484, opened 2016-10-31
Fix hover, autostyle button and dropdown

donflopez:


xavijam:
I miss some tests

donflopez:
Tests added

spec/widgets/dropdown/content-view.spec.js
```diff
@@ -22,6 +22,17 @@ describe('widgets/dropdown/widget-dropdown-view', function () {
...
 
+  it('should not be the options button', function () {
+    $('.js-button').click();
+    expect($('.js-container').find('.js-editWidget').length).toBe(0);
+  });
+
+  it('should be the options button', function () {
+    this.model.set('show_options', true);
+    $('.js-button').click();
+    expect($('.js-container').find('.js-editWidget').length).toBe(1);
```
xavijam: Edit and remove right?

spec/widgets/dropdown/content-view.spec.js
```diff
@@ -22,6 +22,17 @@ describe('widgets/dropdown/widget-dropdown-view', function () {
     expect($('.js-container').find('.CDB-Dropdown').length).toBe(1);
   });
 
+  it('should not be the options button', function () {
+    $('.js-button').click();
+    expect($('.js-container').find('.js-editWidget').length).toBe(0);
```
xavijam: Same here?

src/widgets/widget-model.js
```diff
@@ -58,7 +58,7 @@ module.exports = cdb.core.Model.extend({
   isAutoStyleEnabled: function (autoStyle) {
     var styles = this.get('style');
 
-    if (!styles && (this.get('type') === 'category' || this.get('type') === 'histogram')) return true;
+    if ((!styles || !styles.auto_style) && (this.get('type') === 'category' || this.get('type') === 'histogram')) return true;
```
xavijam: I'd add a test for this.

spec/widgets/dropdown/content-view.spec.js
donflopez: It is not the same, It checks that by default, the options buttons don't appear, and the other just the opposite.

xavijam: Sorry, I don't get it. If options is true, both buttons should appear. If not, they should not appear.

donflopez: Yeah! It's just as you've said! :)

xavijam: So, why don't we check that both buttons are present?

donflopez: Both are on the same conditional I thought it was redundant to check both of them.

xavijam: Why not test there are 4 options instead of 2, not sure.

donflopez: Yes, in that case... I'll check four options, two and if they are the corrects.

donflopez:
Added all of them

PR #485, opened 2016-11-02
Add opts.autoStyle flag

donflopez:
This PR allow deactivate the autostyle button from widgets with the `autoStyle` flag passed as option to createDashboard function.

If this option is not present, the autoStyle feature is disabled by default.


xavijam:
As we have talked offline, we should pass the dashboard model option `autoStyle` to each widget in order to decide if we let the button to be shown.

xavijam:
BTW, branch in builder in order to test this -> `auto-style-feature-flag`

donflopez:
Yep! Let's to add them!

donflopez:
More test added @xavijam, it is ready?

PR #486, opened 2016-11-02
Fix state for histogram

nobuti:
This PR fixes the state issues of #445, related to histogram mainly. 

Basically we are injecting the state form the url, so we need to perform some operations on the widgets when this happens. In the histogram widget it's way more complicated because there are two views acting on the same model. Also some issues related to the stats have been fixed.

This PR is a patch, we need to tackle a better approach for the histogram in the future.

cc @xavijam @javisantana

### DIFF
--- a/spec/widgets/category/content-view.spec.js
+++ b/spec/widgets/category/content-view.spec.js
@@ -10,7 +10,8 @@ describe('widgets/category/content-view', function () {
       column: 'col'
     });
     this.model = new CategoryWidgetModel({
-      title: 'Categories of something'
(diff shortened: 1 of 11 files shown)


nobuti:
I need to fix the tests, but I want to deploy it to staging (ded05) in order to test it as soon as possible.

nobuti:
This is ready for review. Tests have been fixed and the zoom range state removed. Now only open the zoom but not zoomed filter applied.

xavijam:
As @nobuti has said, this is a patch, we should take care of it at length.

PR #489, opened 2016-11-07
Track if dataviewsFetched happened already when adding new widgets.

nobuti:
In order to trigger the bindings properly when adding new widgets, we need to check if the initial
dataviews instantiation has already happened. In this scenario, we need to apply bindings without
waiting this event.

CR @donflopez 

cc @saleiva @javisantana


donflopez:
🇪🇸

javisantana:
I think when we revisit this feature we will need to remove this

nobuti:
Of course, this is a patch, we still need to revamp the approach. Set a widget's initial state is a PITA right now.

PR #497, opened 2016-11-23
Match media for handle breakpoints at js level.

nobuti:
This PR refactor the way we handle actions when the viewport reaches different breakpoints. Also, it introduces the new breakpoints defined in https://github.com/CartoDB/cartodb/pull/10485

CR @hacheka 
cc @xavijam @piensaenpixel

### DIFF
--- a/src/dashboard-sidebar-view.js
+++ b/src/dashboard-sidebar-view.js
@@ -75,11 +75,11 @@ module.exports = cdb.core.View.extend({
   },
 
   _initResize: function () {
-    var mq1280 = matchMedia('(max-width: 1280px)');
-    var mq1600 = matchMedia('(max-width: 1600px)');
-
-    mq1280.addListener(this._resizeHandler);
-    mq1600.addListener(this._resizeHandler);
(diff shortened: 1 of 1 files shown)


xavijam:
Fine by me!

piensaenpixel:
Uhmmm but I think this doesn't fit. The problem with this is that really we have two kinds of responsive view, one for the embed (the actual) and the other one (this PR) for the builder and I think we should keep both.

src/dashboard-sidebar-view.js
```diff
@@ -75,11 +75,11 @@ module.exports = cdb.core.View.extend({
   },
 
   _initResize: function () {
-    var mq1280 = matchMedia('(max-width: 1280px)');
-    var mq1600 = matchMedia('(max-width: 1600px)');
-
-    mq1280.addListener(this._resizeHandler);
-    mq1600.addListener(this._resizeHandler);
+    var breakpoints = ['(max-width: 1600px)', '(max-width: 1280px)','(max-width: 759px)'];
```
nobuti: This is sweet Emilio, I ❤️  it

piensaenpixel:
go!

PR #504, opened 2016-12-28
Represent auto-style colors 

xavijam:
It represents the colors applied with the auto style in the histogram bars.
Related ticket:  https://github.com/CartoDB/cartodb/issues/11095.

screen shot 2017-01-04 at 15 21 05
screen shot 2017-01-04 at 15 24 08
screen shot 2017-01-04 at 15 25 10

**What should you test?**
Having an account with auto-style flag enabled:

- Applying the autostyle over a **histogram widget** should represent the same colors in the histogram bars.
- Disabling autostyle should remove the autostyle colors over the bars.
- Having applied the autostyle, change the number of bins/buckets of the widget and look if the widget is refreshed with the new bins and with the autostyle colors applied over the bars.
- Disable the autostyle and check that bars are back to the default color.
- Apply the autostyle over a histogram widget, check if the colors are ok and then move the map in order to change the widget bars height, color should keep the same.
- Add another histogram widget over the same "layer", and apply autostyle over that new widget. Autostyle colors should be there, ok. Then apply the autostyle for the previous widget in order to disable the autostyle. Newest widget should reset its bars color.
- Check if you change the widget fill color and the histogram is refreshed with that color over the bars.
- Check if you change the widget autostyle colors and they have been represented in the widget bars.
- Add a time-series widget.
- Test all previous changes in the public embed.
- **Last one, but not less important, disable autostyle flag and check that your map and widgets keep working as they should**.

**Available in ded03.**

### DIFF
--- a/src/widgets/histogram/chart.js
+++ b/src/widgets/histogram/chart.js
@@ -4,6 +4,8 @@ var d3 = require('d3');
 var cdb = require('cartodb.js');
 var tinycolor = require('tinycolor2');
 var formatter = require('../../formatter');
+var FILTERED_COLOR = '#1181FB';
+var UNFILTERED_COLOR = 'rgba(0, 0, 0, 0.06)';
 
 module.exports = cdb.core.View.extend({
 
@@ -266,6 +268,8 @@ module.exports = cdb.core.View.extend({
   _onBrushMove: function () {
     this.model.set({ dragging: true });
     this._selectBars();
+    this._setupFillColor();
+    this._refreshBarsColor();
     this._adjustBrushHandles();
   },
 
@@ -295,9 +299,7 @@ module.exports = cdb.core.View.extend({
 
     if (bar && bar.node() && !bar.classed('is-selected')) {
       var left = (barIndex * this.barWidth) + (this.barWidth / 2);
-
       var top = this.yScale(freq);
-
       var h = this.chartHeight() - this.yScale(freq);
 
       if (h < this.options.minimumBarHeight && h > 0) {
@@ -321,7 +323,9 @@ module.exports = cdb.core.View.extend({
       .attr('fill', this._getFillColor.bind(this));
 
     if (bar && bar.node()) {
-      bar.attr('fill', this._getHoverFillColor.bind(this));
+      bar.attr('fill', function (d, i) {
+        return this._getHoverFillColor(data[barIndex], barIndex);
+      }.bind(this));
       bar.classed('is-highlighted', true);
     }
   },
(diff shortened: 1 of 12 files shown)


saleiva:
Quick thing detected in pre-acceptance ;)

When auto-style is selected and we filter the data, bars should be colored with the gradient instead of blue.

src/widgets/histogram/chart.js
```diff
@@ -1021,8 +1077,11 @@ module.exports = cdb.core.View.extend({
       .enter()
       .append('rect')
       .attr('class', 'CDB-Chart-bar')
-      .attr('fill', this.options.chartBarColor)
-      .attr('data', function (d) { return _.isEmpty(d) ? 0 : d.freq; })
+      .attr('fill', this._getFillColor.bind(self))
+      .attr('data', function (d) {
```
nobuti: this chunck is being repeated above.

src/widgets/widget-model.js
```diff
@@ -72,8 +72,14 @@ module.exports = cdb.core.Model.extend({
 
     var styles = this.get('style');
 
-    if ((!styles || !styles.auto_style) && (this.get('type') === 'category' || this.get('type') === 'histogram')) return true;
-    return styles && styles.auto_style && styles.auto_style.allowed;
+    if (this.get('type') === 'category' || this.get('type') === 'histogram') {
+      if (!styles || !styles.auto_style) {
```
nobuti: not sure if we can short this as `return styles && styles.auto_style && styles.auto_style.allowed || true`

src/widgets/histogram/chart.js
```diff
@@ -940,6 +950,46 @@ module.exports = cdb.core.View.extend({
...
+        autoStyleColorsScale = d3.scale.quantize()
+          .domain(domain)
+          .range(colorsRange);
+      }
+
+      this._autoStyleColorsScale = autoStyleColorsScale;
+    }
+  },
+
+  _getFillColor: function (d, i) {
```
ivanmalagon: Just being picky here but... can we use more descriptive argument names than `d`?

Argument `i` doesn't seem to be used in the function, as well.

src/widgets/widget-model.js
xavijam: :+1:

matallo:
LGTM, awesome work with the `chart.spec.js` specs (and the `chart.js` ofc) 👍

xavijam:
Doing acceptance as well with @javitonino.

javitonino:
Looking good, I could only find a couple of minor issues:
 - You can't change the fill color of a timeline widget you just added. The widget color is not updated. After refreshing the builder, this works as expected; it only happens just after creating the widget.
 - On Firefox/Linux, in the embed, if you autostyle by a histogram, and then filter by a different histogram, the bars in the autostyles widget turn transparent (but you can still interact with them).

xavijam:
Let me test it in IE11 and check the error with the time-series.

xavijam:
After testing several browsers:

- Chrome (last version): Mac -> ✅ 
- Firefox (last version): Mac -> ✅ 
- Opera (last version): Mac -> ✅ 
- Safari (last version): Mac -> ✅ 
- IE11 (last version): Win7 -> ✅ 

Pending Ubuntu, gimme some minutes, but I feel we can go on.

javitonino:
Another way to reproduce the invisible bars which works in Chrome & Firefox. In the builder, add two histograms, autosytle by one and filter by the other. Then refresh the builder and the bars are transparent when the state is restored.

xavijam:
Deploying last changes in ded03 @javitonino, if you have the chance to test it under your environment, it would be great :D.

javitonino:
The issue in Firefox+Chrome in the builder is solved 🎉. The issue that only affected Firefox/Linux in the embed is still there though.

I also did a quick pass through everything and nothing broke that I could see.

xavijam:
Merging and let's check the other problem afterwards 💋 .

PR #506, opened 2017-01-02
Don't update auto-style colors when the feauture is disabled

xavijam:
This problem is related with https://github.com/CartoDB/cartodb/issues/11166. We can't update auto-style colors when the feature is disabled. That throws an error and makes the app fail.

**STT**
- Basically the same steps described in https://github.com/CartoDB/cartodb/issues/11166.

Uploading to ded02 in order to be tested.


xavijam:
Deploying changes in ded02.

matallo:
LGTM, I'd ask why not testing `_updateColors` to have been called but I suppose that is covered with the `autoStyle()` or `colors.updateData()`

xavijam:
yeah, rest of scenarios are already covered, so 👍 .

MariaCheca:
Acceptance passing 🍪

PR #512, opened 2017-01-16
11246 issue

xavijam:
Ticket: cartodb/cartodb#11246.

Changing way to escape quotes with auto-style, following the same way we use within Builder and categories.

**Things to test:**
Check the ticket, because there is a good example there, but in any case:
- Create a map with a layer where the data contains single and double quotes.
- Add a category widget using the column where those single and double quotes are present.
- Autostyle that widget.


MariaCheca:
Acceptance passing 🇩🇰

Issue #513, opened 2017-01-25
Autostyle button should not be enabled if autostyler can't provide a colors range 


xavijam:
Original ticket: https://github.com/CartoDB/cartodb-management/issues/4815.

Let me show the problem with an example:

**STEPS TO REPRODUCE IT (BUILDER)**
- Create a map with a layer of points.
- Take a column where there are different values, it doesn't matter if it is string or number.
- In my case I'll take a number column called "count".
- Add a new histogram widget, using that "count" column.
- Go to the style tab, and add a custom style. It could be:

```css
#layer {
   marker-fill-opacity: 0.9;
   marker-line-color: #FFF;
   marker-line-width: 1;
   marker-line-opacity: 1;
   marker-placement: point;
   marker-type: ellipse;
   marker-width: 20;
   marker-allow-overlap: true;
}

#layer[count=100] {
   marker-file: url(ANYURL);
}
#layer[count<100] {
   marker-file: url(ANYURL);
}
```
- The important part is that there is a marker-file without any marker-fill.
- Refresh the page.
- 💥 .

This is due to we try to get the autostyle colors in order to pre-generate the ramps but there are not colors provided from the AutoStyler, so d3 interpolation fails. That's the first problem, but easy to solve with a `return` when `colorsRange` is undefined.

- If you fix that ^^^, refresh the page and check that the map is working again.
- Click now in the autostyle button.
- 💥 .

We should not display the autostyle button because there is no possibility to autostyle anything, the style is custom and there are icons/images, not "simple" colors. We should hide the button in this case.

cc @nobuti

javitonino:
The autostyle button is not hidden for category widgets.

xavijam:
In production!

PR #514, opened 2017-01-25
Don't offer auto-style if it is not possible to auto-style due to a custom cartocss with image

xavijam:
Problem coming from: https://github.com/CartoDB/cartodb-management/issues/4815, and reproduced here https://github.com/CartoDB/deep-insights.js/issues/513.

**THINGS TO TEST**

- Apply a custom CartoCSS with icons (not common circles or anything), check that autostyle button disappears.
- Unapply custom CartoCSS, apply a standard one and check that autostyle button is back.
- Check that other combinations are not affected, like testing it with layer visibility, if it is allowed, etc.
- Check that, if a histogram widget is added when the cartocss doesn't provide any [marker-fill or polygon-fill or line-color], autostyle button doesn't appear. After that, reset the cartocss, provide a simple one, and autostyle button should appear again, click it and check that bars appear colored properly.
- Check time-series widget could be added correctly without problems.

### DIFF
--- a/spec/widgets/category/search-title-view.spec.js
+++ b/spec/widgets/category/search-title-view.spec.js
@@ -23,6 +23,7 @@ describe('widgets/category/search-title-view', function () {
       this.widgetModel = new CategoryWidgetModel({}, {
         dataviewModel: this.dataviewModel
       }, {autoStyleEnabled: true});
+
       this.view = new SearchTitleView({
         widgetModel: this.widgetModel,
         dataviewModel: this.dataviewModel
@@ -35,7 +36,6 @@ describe('widgets/category/search-title-view', function () {
       expect($el.find('.js-title').length).toBe(1);
       expect($el.find('.CDB-Widget-options').length).toBe(1);
       expect($el.find('.js-titleText').length).toBe(1);
-      expect(this.view.$('.js-autoStyle').length).toBe(1);
     });
 
     describe('search', function () {
@@ -87,6 +87,7 @@ describe('widgets/category/search-title-view', function () {
 
     describe('options', function () {
       beforeEach(function () {
+        spyOn(this.view, '_isAutoStyleButtonVisible').and.returnValue(true);
         this.view.render();
       });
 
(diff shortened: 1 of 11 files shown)


xavijam:
Deploying to ded04 + creating PR in BUILDER.

xavijam:
PR in builder: https://github.com/CartoDB/cartodb/pull/11350

spec/widgets/category/search-title-view.spec.js
```diff
@@ -35,7 +36,6 @@ describe('widgets/category/search-title-view', function () {
       expect($el.find('.js-title').length).toBe(1);
       expect($el.find('.CDB-Widget-options').length).toBe(1);
       expect($el.find('.js-titleText').length).toBe(1);
-      expect(this.view.$('.js-autoStyle').length).toBe(1);
```
javitonino: ~~Suggestion: Maybe it's better to stub the `_isAutoStyleButtonVisible` (like below) rather than removing this check?~~

Edit: Disregard, this is tested in depth below.

xavijam: Yep! :es:

javitonino:
Seems to work fine if you keep yourself to category and histogram widgets. But I can reproduce an issue when adding time series into the mix:
1. Create a points map
2. Add a histogram widget for cartodb_id
3. Add a time-series for cartodb_id

The second widget will not load correctly. If you refresh, it also won't load one of the widgets (appears to depend on widget ordering).

javitonino:
Updated deps in Builder branch and deploying to ded04 for testing.

javitonino:
I hate to be :thatguy: but I got to get this back to development due to category widgets. 

STR: Create a map (default styles are fine) and add a category widget. It doesn't have the autostyle button. If you change the style, the button reappears, so it seems like an initialization issue.

Note, it may be a race condition. In a couple of occasions, the category widget autostyle worked right after adding it, but in the inmense majority, it did not.

xavijam:
Fixed a last problem with histogram charts, added the thing to test at the beginning.

javitonino:
Ok, so I've been testing this for a while in staging and it seems to work pretty well. Histograms widget correctly hide the autostyle button when needed, and category widgets keep it always on (although sometimes it may do nothing).

There is only an edge case I found that breaks this. We should decide whether to address this in this PR, or report it as a follow-up ticket. The STR is as follow:
1. Create a point map
2. Apply an style for a polygon (you see where this is going). For example: `#layer { polygon-fill: #FFB927; }
3. Add an histogram widget
4. Click autostyle
5. :boom: goes the dynamite

As I said, this is an edge case as it only triggers when you manually apply styles for the incorrect type of geometry.

P.S: I just tested and this happens in production as well. So I'd deploy this and report this as a new ticket. What do you think?

javitonino:
We agreed offline to merge this and split the other issue to https://github.com/CartoDB/cartodb/issues/11363

PR #515, opened 2017-01-25
change of classification method for F shape distribution

makella:
This is a change to the classification method we use behind the scenes for the Histogram widget with a F shaped distribution. I had originally set it to quantiles but after more thought, equal is more appropriate.

Please let me know if you have any questions

cc @saleiva


xavijam:
@donflopez could you take a look please?

xavijam:
@makella, should we merge it then?

makella:
yes, please!

xavijam:
Adding to our kanban in order to deploy it on monday 💋

nobuti:
To be deployed as part of https://github.com/CartoDB/cartodb/pull/11639

PR #516, opened 2017-01-27
Don't provide autostyle definition if autostyler variable is not defined or if auto style is not enabled...

xavijam:
cc @ivanmalagon

### DIFF
--- a/spec/widgets/widget-model.spec.js
+++ b/spec/widgets/widget-model.spec.js
@@ -190,6 +190,21 @@ describe('widgets/widget-model', function () {
       });
     });
 
+    describe('.getAutoStyle', function () {
+      it('should not provide any info if auto-style is not enabled', function () {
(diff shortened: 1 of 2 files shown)


spec/widgets/widget-model.spec.js
```diff
@@ -190,6 +190,21 @@ describe('widgets/widget-model', function () {
...
+      });
+
+      it('should not provide any info if autoStyler is not defined', function () {
+        spyOn(this.model, 'isAutoStyleEnabled').and.returnValue(true);
+        this.model.autoStyler = undefined;
+        var data = this.model.getAutoStyle();
+        expect(data).toEqual({});
+      });
+    });
+
```
ivanmalagon: This spec only test one path, the one where we shouldn't provide any info.

There was no existing specs about the other case, where we should provide info, but I think this is the right moment to add them.

xavijam:
Reopening ticket until it is finally merged.

Issue #385, opened 2016-06-27
Mobile issues in the public view


saleiva:
- [x] Interaction in mobile is pretty hard. Tested on Android
- [x] Keyboard appears and disappears when you click on search on category widget

saleiva:
Prioritizing since more maps are being published with widgets

nobuti:
Could you list the troubles you found?

saleiva:
Yep, 

Filtering is very hard (you have very low precission). 
On the other hand, once you've filtered, moving the lateral bars is almost impossible.

jorgesancha:
Moving to NEXT to prioritise Style with Icons and some backend stuff cc @saleiva

piensaenpixel:
@nobuti maybe the problem with the scroll could be fixed in CSS, but I need to check it

piensaenpixel:
This is the most important problem:

![widgets](https://cloud.githubusercontent.com/assets/1214201/22796321/4ad98fde-eefa-11e6-98bb-34343002e9f3.gif)

It's impossible to filter in the histogram because the scroll's event intercept the filter

jorgesancha:
thanks @piensaenpixel 👏

noguerol:
So, just for giving a little more of context: we need this for a demo we are building for the Telefonica's booth in Mobile World Congress. MWC will start in two weeks, but the deadline is even tighter because next Wed 15 is the last day we can test this on the Surface device that will run this map.

nobuti:
@saleiva @xavijam @jorgesancha @piensaenpixel we have a problem with the category widget and the search thing. Doing focus on the field triggers the keyboard to appear. So the viewport is resized. When resized the dataviews change, so all the widgets views are rendered again. So the field that had got the focus doesn't exist anymore, so the focus is gone, and the keyboard hides. And the cycle begins again. Basically, this feature is unusable in mobile.

nobuti:
For the time being, I'm going to hide the search button in devices with touch support.

xavijam:
Not sure, what could we do better here?

nobuti:
Honestly, the changes involved in order to fix it are too deeper to consider right now. Given that the user is able to filter categories even without searching, it's not that bad.

xavijam:
:+1:

xavijam:
Deployed in production.

noguerol:
I agree with the decision, but let's start thinking on a more long term solution: https://github.com/CartoDB/deep-insights.js/issues/520

PR #519, opened 2017-02-13
Fixes.

nobuti:
This PR implements #385.

- It hides search button on mobile. Right now the search functionality is impossible to use in mobile because the keyboard resizes the viewport. Then all views are rendered again, and finally the keyboard hides.
- It adds proper class to let ps know it does not have to handle the `touchmove` event. This depends on https://github.com/CartoDB/perfect-scrollbar/pull/1.

Besides this PR and the one commented above, we need to update cartodb.js as well (there is already a branch for that). The order to merge should be: PS repo, cartodb.js and the last one, this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,12 +24,12 @@
   "dependencies": {
     "backbone": "1.2.3",
     "cartocolor": "4.0.0",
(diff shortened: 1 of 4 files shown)


xavijam:
For me it works :+1:

PR #521, opened 2017-02-15
Fixing problem related with Surface map pan

xavijam:
Same problem described for the map [here](https://github.com/CartoDB/cartodb.js/pull/1580) but happening with chart widgets (histogram and time-series).

cc @nobuti @piensaenpixel @alonsogarciapablo

### DIFF
--- a/themes/scss/widgets/default.scss
+++ b/themes/scss/widgets/default.scss
@@ -110,6 +110,8 @@
   .CDB-Widget-content--timeSeries {
     width: 100%;
     margin-top: 0; // override default
+    touch-action: none;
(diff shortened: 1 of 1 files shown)


themes/scss/widgets/default.scss
```diff
@@ -77,6 +77,12 @@
   margin-right: -7px;
 }
 
+.CDB-Widget-content--histogram,
+.CDB-Widget-content--timeSeries {
+  touch-action: none;
+  -ms-touch-action: none;
```
xavijam: Let me do it hound 💋

PR #524, opened 2017-02-24
Removed unnecessary nice function from time-series header view

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11174

We were rounding (nicely) selection dates, and it was causing problems in the way we calculate selected ranges. 

An example with the same dataset, before and after:

Before:
screen shot 2017-02-24 at 13 01 58

After:
screen shot 2017-02-24 at 13 02 21

cc @michellechandra  @zingbot


nobuti:
🚀

PR #527, opened 2017-03-09
Jquery sizzle weird silent bug.

nobuti:
See http://stackoverflow.com/questions/21661113/jquery-error-during-load-failed-to-execute-query-x-is-not-a-valid-selecto for details.

cc @xavijam


xavijam:
@CartoDB/frontend, we have discovered this problem after diving into a TrackJS notification. It is a bit WTF because the error itself is not being caught.

xavijam:
Basically, we have two options:

1. Remove jquery (ponte ahora con eso :joy:).
2. Don't make use of `:eq(x)`, `:first-child`, ... typical CSS selectors with jQuery.

PR #528, opened 2017-03-09
fix loader in low resolutions view

piensaenpixel:
- [X] Change loader's size in widgets and also change size in low screen's resolutions
- [X] Fix overflow in low screen's resolutions


xavijam:
Tested and working! 👍

PR #531, opened 2017-03-21
When dataBin is not defined it should not return anything

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/11798

Steps to reproduce already described there ^^^. But good ones also are:

- Create a map with a histogram widget where the there is at least a bucket without data.
Example: 
screen shot 2017-03-21 at 19 12 10

- Select the bin or bins without data and zoom into it/them.
Example:
screen shot 2017-03-21 at 19 12 15
- JS error appeared, now zoomed histogram should be empty.


nobuti:
Tested, looking good 👍

PR #532, opened 2017-03-27
Pointing to new version of CartoDB.js with export image feature

xavijam:


xavijam:
Not necessary anymore.

PR #533, opened 2017-04-03
Replace way to trigger widget state changes

xavijam:
We have agreed a better way to "report" when a state change has happened. So any other elements can subscribe to it.

Related PR in CARTO Builder: https://github.com/CartoDB/cartodb/pull/11901#pullrequestreview-30460811

### DIFF
--- a/src/api/dashboard.js
+++ b/src/api/dashboard.js
@@ -64,7 +64,11 @@ Dashboard.prototype = {
   },
 
   _widgetsAlreadyInitialized: function () {
-    return this._dashboard.widgets._widgetsCollection.hasInitialState();
+    var widgetsCollection = this._dashboard.widgets._widgetsCollection;
+    if (widgetsCollection.size() > 0) {
+      return this._dashboard.widgets._widgetsCollection.hasInitialState();
+    }
+    return true;
   },
 
   onDataviewsFetched: function (callback) {
@@ -74,7 +78,7 @@ Dashboard.prototype = {
     } else {
       this._dashboard.vis.once('dataviewsFetched', function () {
         callback && callback();
-      }, this);
+      });
     }
   },
 
--- a/src/api/dashboard.js
+++ b/src/api/dashboard.js
@@ -3,6 +3,13 @@ var _ = require('underscore');
 
 function Dashboard (dashboard) {
   this._dashboard = dashboard;
+
+  this.onDataviewsFetched(function () {
+    dashboard.widgets._widgetsCollection.initialState();
+    dashboard.widgets._widgetsCollection.each(function (m) {
+      m.applyInitialState();
+    });
+  });
 }
 
 Dashboard.prototype = {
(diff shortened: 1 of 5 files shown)


xavijam:
@nobuti @alonsogarciapablo

src/api/dashboard.js
```diff
@@ -56,14 +63,21 @@ Dashboard.prototype = {
...
+    } else {
+      this._dashboard.vis.once('dataviewsFetched', function () {
+        callback && callback();
       }, this);
-      shareURLs === true && this._bindChange(callback);
-    }, this);
+    }
+  },
+
+  onStateChanged: function (callback) {
```
alonsogarciapablo: ✨

xavijam:
Could you take a look guyis?

src/api/dashboard.js
```diff
@@ -56,14 +63,29 @@ Dashboard.prototype = {
...
-  onStateChanged: function (callback, shareURLs) {
-    this._dashboard.vis.once('dataviewsFetched', function () {
-      this._dashboard.widgets._widgetsCollection.initialState();
-      this._dashboard.widgets._widgetsCollection.each(function (m) {
-        m.applyInitialState();
-      }, this);
-      shareURLs === true && this._bindChange(callback);
-    }, this);
+  _widgetsAlreadyInitialized: function () {
+    var widgetsCollection = this._dashboard.widgets._widgetsCollection;
```
xavijam: I have added this last thing because testing it in Builder I noticed that when there are no widgets added, callback was not triggered. Thanks acceptance you exist :joy:.

alonsogarciapablo: If we could avoid this kind of 🚋 🚋 🚋  it'd be 🔝 . We're breaking the Law of Demeter & we're accessing private attributes, which might change in the future and break this...  We could just ask `this_dashboard` if it has widgets and they are initialised: 

```
if (this._dashboard.areWidgetsInitialised()) {
 // Do wathever
}
```

I know `this._dashboard` is not a "real" object so we might need to convert it into one.

src/api/dashboard.js
```diff
@@ -56,14 +63,29 @@ Dashboard.prototype = {
...
+    }
+    return true;
+  },
+
+  onDataviewsFetched: function (callback) {
+    var areDataviewsFetched = this._widgetsAlreadyInitialized();
+    if (areDataviewsFetched) {
+      callback && callback();
+    } else {
+      this._dashboard.vis.once('dataviewsFetched', function () {
```
alonsogarciapablo: We could even move this method to CartoDB.js since it's basically asking about dataviews (a concept that belongs to CartoDB.js):

```
onStateChanged: function (callback) {
  if (this._dashboard.vis.onDataviewsFetched(callback)) {
    // something....
  });
},
```

src/api/dashboard.js
xavijam: It will need changes in CartoDB.js. Then we will request changes in Javascript v8 engine, won't we?.

src/api/dashboard.js
xavijam: Agree, but I am not sure about making this issue bigger and bigger...

src/api/dashboard.js
xavijam: Forget my previous comment, we could provide it but it should be added in the object returned when dashboard is created, do you think it is worth it?

xavijam: Agree let me show you the solution.

nobuti:
This opens the 🍈 to export the anonymous mapView, just saying.

src/api/dashboard.js
xavijam: Well, not that agree. If we make this change we could end up in some confusions. In DI we only need the first event triggered, but in CARTO.js we are sending the same event when all dataviews are fetched. So that means the "binding" should be decided in the DI field (bind, once, stop, off, on, whatever we want)... I'd leave it for now as it is.

xavijam:
Really @nobuti ? 🤔

Issue #518, opened 2017-02-03
Polygon lines when auto-style is applied turn the lines black


donflopez:
This is a problem with the polygons lines because auto-style doesn't take into account the kind of features that a layer has, and apply the ramp or whatever to all of the shapes, so the lines turn black (the default colour)

Steps:

Apply auto-style in some map with polygons

Result:

Black polygons lines

Expected:

Polygon lines with the original colour

Comments:

This is an error reported [here](https://github.com/CartoDB/tangram-carto/issues/29).

cc\ @saleiva

makella:
I did a little digging and you guys probably already know this, but this is what I was seeing just to document so I can reference with others until we have a fix.

The issue is that `line-color` is getting ramped in addition to `polygon-fill`

on a category widget the cartocss that is generated from auto-style looks like this


```
#layer ['mapnik::geometry_type'=2] { line-color: ramp([name], ("#7F3C8D", "#11A579", "#3969AC", "#F2B701", "#E73F74", "#A5AA99"), ("RM13", "BS39", "M12", "SO16", "TD5"));  } #layer ['mapnik::geometry_type'=3] { polygon-fill: ramp([name], ("#7F3C8D", "#11A579", "#3969AC", "#F2B701", "#E73F74", "#A5AA99"), ("RM13", "BS39", "M12", "SO16", "TD5"));  }  

#layer {
 
  polygon-opacity: 0.9;
  polygon-gamma: 0.5;
 
  line-width: 1;
  line-opacity: 0.5;
  line-comp-op: soft-light;
}
```
the culprit is the `#layer ['mapnik::geometry_type'=2] { line-color:` since that doesn’t ‘work’ it applies the default line color which is black hence the black lines. You'll notice that if you remove the `['mapnik::geometry_type'=2]` piece that things are fixed.

Possible workaround (for now):
- make the line-width and line-opacity `0.25`
and/or
- remove the line properties completely

makella:
I also want to note that I went through each of the 'advanced' styling things like `polygon-gamma: 0.5;` and `line-comp-op: soft-light;`.

Even if those are removed, we get the same result because of the `['mapnik::geometry_type'=2]`.

I just wanted to make sure it wasn't my fault! Haha :)

donflopez:
Thank you @makella! You're completely right.

I just wanted to add that, tangram takes the lines of the polygons as polygon lines, so... his geometry is polygon, not lines...

[...]

PR #537, opened 2017-04-17
Fix AS lines with vector

donflopez:
#518


xavijam:
Taking care of it in https://github.com/CartoDB/deep-insights.js/pull/558.

PR #552, opened 2017-06-21
added source to widgets

matallo:
re: https://github.com/CartoDB/cartodb/issues/9956

list of analyses [1] and i18n implementation [2] as well as layer-colors and numbers [3] have been blatantly copied from CartoDB/cartodb

after discussion with @xavijam `sourceId`, `sourceType`, and `layerName` come from the dataviewModel, and `sourceColor` has been delegated to the widgetModel, replicating the same behaviour than in Builder, where a `letter` and `color` attributes are added to layerDefinitionModel

we use the `show_source` in the widgetModel to control the visibility, and they won't be visible by default (public views), however in Builder dashboards they will appear, controlled with deep insights integration

[1] https://github.com/CartoDB/deep-insights.js/blob/f7e3f95b91b3b17f23e32adf19e2bc0c4d6e971b/src/data/analyses.js
[2] https://github.com/CartoDB/deep-insights.js/blob/3adb9db40608e656d5418d3306e573f87dfd61d6/src/index.js#L1-L8
[3] https://github.com/CartoDB/deep-insights.js/blob/3adb9db40608e656d5418d3306e573f87dfd61d6/src/util/layer-colors.js


package.json
```diff
@@ -18,17 +18,19 @@
     "Raul Ochoa <rochoa@cartodb.com>",
     "Nicklas Gummesson <nicklas@cartodb.com>",
     "Pablo Alonso <pablo@cartodb.com>",
-    "Antonio Santiago Pérez <asantiago@cartodb.com>"
+    "Antonio Santiago Pérez <asantiago@cartodb.com>",
+    "Carlos Matallín <matallo@carto.com>"
```
nobuti: 🙊

xavijam: ... come on! let's remove the contributors please! :joy:
It appears in the Github tab.

matallo: that's my first change in any PR 😬

src/data/analyses.js
```diff
@@ -0,0 +1,121 @@
+var MAP = {
```
nobuti: we should extract this to a common place, not sure how. I'm worried about adding a new analysis and forgetting to update this.

src/widgets/formula/content-view.js
```diff
@@ -47,18 +49,27 @@ module.exports = cdb.core.View.extend({
...
 
+    var sourceId = this._dataviewModel.get('source').id;
+    var letter = layerColors.letter(sourceId);
+    var sourceColor = layerColors.getColorForLetter(letter);
+
     this.$el.html(
       template({
         title: this.model.get('title'),
+        sourceId: sourceId,
+        sourceType: analyses.title(this._dataviewModel.getSourceType()),
```
nobuti: I'd add `|| ''` if it doesn't exist, I guess it could prevent to show an `undefined`

matallo: if it doesn't exist the `sourceType` is actually "source" and comes from cartodb.js, but I added it anyway

src/data/analyses.js
matallo: in the worst case (the key not existing because it has not been updated) a non-interpolated string would appear, from `.getAnalysisByType`, that's the case for "Source" analyses

it's difficult to decouple from cartodb, because in the `data/analyses.js` there's stuff regarding backbone forms, for example https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/data/analyses.js#L16

what would be your proposal?

matallo:
you're totally right @nobuti, added!

also, I moved the `_t` becuase it was messing with the `window._t` object in Builder

src/data/analyses.js
nobuti: Right, the only thing I can think of it's a npm module, but you are right, it's not the time.

PR #554, opened 2017-06-23
Range filter tags + Widget title

rjimenezda:
Fix https://github.com/CartoDB/cartodb/issues/12291
Fix https://github.com/CartoDB/cartodb/issues/12178

This pull request:

- Changes how range handles look
- Changes how axis tips look
- Adds widget title to Time Series

asdf

The responsive design is slightly different, and requires extra work, but in the meantime I've attempted to leave the mobile version usable.


rjimenezda:
There's a minor change required in cartodb, essentially updating the styles for the context switcher buttons, since widgets are a bit taller now. I'll cross reference the PRs

themes/scss/widgets/themes/_light.scss
```diff
@@ -216,8 +216,20 @@ $chartMiniSelected: rgba(#333, 1);
     stroke-width: 2;
     fill: $commonBkg;
   }
+  .CDB-Chart-axisTipRect {
+    fill: #eeeeee;
+  }
+  .CDB-Chart-axisTipRect--inverted {
```
rjimenezda: I hate OSX sometimes 🙃 🔫

src/widgets/histogram/content-view.js
```diff
@@ -332,7 +333,7 @@ module.exports = cdb.core.View.extend({
     var $tooltip = this.$('.js-tooltip');
 
     if (info && info.data) {
-      var bottom = this.defaults.chartHeight + 3 - info.top;
+      var bottom = this.defaults.chartHeight + 23 - info.top;
```
nobuti: Why this magic number? I'd change it by a constant or something to clearly represent the intention of the code.

themes/scss/widgets/themes/_light.scss
```diff
@@ -216,8 +216,20 @@ $chartMiniSelected: rgba(#333, 1);
     stroke-width: 2;
     fill: $commonBkg;
   }
+  .CDB-Chart-axisTipRect {
+    fill: #eee;
```
nobuti: Yep.

themes/scss/widgets/themes/_light.scss
```diff
@@ -216,8 +216,20 @@ $chartMiniSelected: rgba(#333, 1);
...
+  }
+  .CDB-Chart-axisTipRect--inverted {
+    font-weight: bold;
+    fill: $primary;
+  }
   .CDB-Chart-axisTipText {
-    fill: $link;
+    font-size: 11px;
```
nobuti: AFAIK we don't have any 11px font size in Builder, is that correct @piensaenpixel?

src/widgets/histogram/content-view.js
rjimenezda: I thought I removed them all, I'll give a 👀

themes/scss/widgets/themes/_light.scss
rjimenezda: Oh I actually fixed this on friday but it's unpushed 🤦‍♂️

themes/scss/widgets/themes/_light.scss
rjimenezda: I probably did that while drafting the new tooltip style, I'll see what the class should be to avoid adding an extra font-size there.

xavijam:
Apart from the @nobuti comments, we will need @CartoDB/design involved here.

nobuti:
It works really well, the tooltip's arrow redraw is ver smooth, impressive. I found we missed something in the way. Now that we have title, the histogram's height is bigger, perhaps we should change the placeholder's height too to avoid the big jump of context once the widget is ready. Right now the difference is too big IMO.

xavijam:
You are right @nobuti, also I guess we will need to change the position of the "geometry edition" buttons and so on.

rjimenezda:
@xavijam Context switching icons CSS is changed in Cartodb (https://github.com/CartoDB/cartodb/pull/12382), but I didn't take into consideration the placeholder version. Will give it a look

xavijam:
> @xavijam Context switching icons CSS is changed in Cartodb (CartoDB/cartodb#12382), but I didn't take into consideration the placeholder version. Will give it a look

Masters!

nobuti:
@CartoDB/design any comment? In ded02.

xavijam:
Checked in Firefox, working smooth too, great job @rjimenezda. My comments:

- For me (IMHO), handles are to wide, but it is likely it works better than a normal ones (you guys decide 💋 ).
- Tooltip ticks are higher than expected, I feel (picky thing #1).

screen shot 2017-06-26 at 11 48 04

- When tooltip is at the end (it works perfect), I feel tick should be positionated to the left|right. It would be like a nice touch (picky thing #2).

nice-to-have

- And now, a question. Should we let the user move the handlers from the "tooltip"?

xavijam:
By the way, I feel we will need to review the heights and so on when source node info is added within any widget.

josecruz:
Testing this guys 💃

rjimenezda:
I've pushed a couple of fixes for the review comments @nobuti 

It also makes the tooltips closer to the handles:

screen shot 2017-06-26 at 12 23 40

I like the subtle triangle change when on the right / left, I'll see how hard it is, I can probably change the path.

The handle size is a parameter, so I'll leave it to design. The bigger it is, the easier it is to grab (:raur:), but I think we can go a bit smaller.

josecruz:
> When the tooltip is at the end (it works perfectly), I feel tick should be positioned to the left|right. It would be a nice touch (picky thing #2).

^ Would be awesome, super nice detail @xavijam! 

> For me (IMHO), handles are too wide, but it is likely it works better than normal ones (you guys decide 💋 ).

Agree but it's true that they are easier to use

> Tooltip ticks are higher than expected, I feel (picky thing #1).

+1

> The histogram is too small on small devices, isn't it? @urbanphes

+1

And we also found some things:

- Handles stroke should be 1px like in the design. Also, the lines inside looks a bit wider (should be 4px wide). And one last picky thing, the tooltips corners looks a bit different as well, maybe the border radius?:
image


- When you add a TS widget directly from the Widget tab, it works great, but if you do it inside a layer, you get this (where right corner buttons seems to be sticked there):
kapture 2017-06-26 at 12 32 36

rjimenezda:
@josecruz will give those a look, thanks for the feedback! The last case should be fixed but I'll check it nevertheless.

By the way, I was trying to center the tooltip with the handle, but it occured to me we're free to position it to, for instance, a fifth to the right / left:

screen shot 2017-06-26 at 12 56 57

Whatever you like the most 👍

rjimenezda:
Changes are in ded01 again @josecruz @xavijam

xavijam:
I've noticed the "current" tick element a bit out of the histogram when it is at the end.

screen shot 2017-06-28 at 10 09 38

Is it related to any of our changes or happening in production?

rjimenezda:
I can't reproduce that, they seem to behave the same way.

Any tips to get it to happen?

xavijam:
> Any tips to get it to happen?

xavijam:
By the way, what about moving the handlers with the tooltips @rjimenezda ?

rjimenezda:
It seems to happen on production as well, it's very subtle though, and only in very specific cases 🤔

xavijam:
> It seems to happen on production as well, it's very subtle though, and only in very specific cases 🤔

👍 

What about the tests failing?

rjimenezda:
retest this please

rjimenezda:
It's a cartodb.js issue it seems

`>> Error: Cannot find module 'babelify' from '/home/travis/build/CartoDB/deep-insights.js/node_modules/cartodb.js/node_modules/tangram.cartodb'`

josecruz:
Hello guys!

From the design point of view, it looks perfect!! 

Great job team ✨  ❤️

src/widgets-service.js
```diff
@@ -163,7 +163,7 @@ WidgetsService.prototype.createTimeSeriesModel = function (attrs, layer, state,
   attrs.column_type = attrs.column_type || 'date';
   var dataviewModel = this._dataviews.createHistogramModel(layer, attrs);
 
-  var attrsNames = ['id', 'style', 'normalized'];
+  var attrsNames = ['id', 'style', 'title', 'normalized'];
```
nobuti: IMO if this is a "constant", we should treat it following the conventions for that, this is uppercased.

themes/scss/widgets/_chart.scss
```diff
@@ -6,7 +6,7 @@
   @include justify-content(space-between, justify);
   @include align-items(flex-end);
   width: 100%;
-  height: 75px; // match real widget content height, to avoid "jumpy" behavior
+  height: 107px; // match real widget content height, to avoid "jumpy" behavior
```
nobuti: how does this behave being responsive? any chance that the title goes in two lines?

themes/scss/widgets/themes/_light.scss
```diff
@@ -216,8 +216,20 @@ $chartMiniSelected: rgba(#333, 1);
...
+  .CDB-Chart-axisTipRect {
+    fill: #eeeeee;
+  }
+  .CDB-Chart-axisTipRect--inverted {
+    font-weight: bold;
+    fill: $primary;
+  }
   .CDB-Chart-axisTipText {
-    fill: $link;
+    font-weight: bold;
```
nobuti: Yep

src/widgets-service.js
rjimenezda: You mean the array, right? Sure 👌

themes/scss/widgets/themes/_light.scss
rjimenezda: There's no font-size though 🤔

themes/scss/widgets/_chart.scss
rjimenezda: Good catch, I'll apply ellipsis to the title to prevent the jumps 👍

PR #556, opened 2017-06-29
widget color changed

matallo:
re: https://github.com/CartoDB/cartodb/issues/12394

adding `widget_color_changed` attribute to the widget definition model style, because otherwise when selecting the category in the widget, it prevails as the widget color is persisted

I didn't quite like the hardcoded color check but I've seen it in other parts of deep insights codebase. Another workaround would be running a migration in the widgets table to add the `widget_color_changed` when the widget color won't match the default color

### DIFF
--- a/src/widgets/widget-model.js
+++ b/src/widgets/widget-model.js
@@ -89,6 +89,7 @@ module.exports = cdb.core.Model.extend({
     var styles = this.get('style');
 
     return styles && styles.widget_style &&
+          styles.widget_style.widget_color_changed &&
           styles.widget_style.definition &&
           styles.widget_style.definition.color &&
           styles.widget_style.definition.color.fixed;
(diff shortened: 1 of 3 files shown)


src/widgets/widget-model.js
```diff
@@ -87,11 +87,14 @@ module.exports = cdb.core.Model.extend({
...
     var styles = this.get('style');
-
-    return styles && styles.widget_style &&
-          styles.widget_style.definition &&
-          styles.widget_style.definition.color &&
-          styles.widget_style.definition.color.fixed;
+    var widgetStyle = styles && styles.widget_style;
+    var widgetColor = widgetStyle && widgetStyle.definition &&
+      widgetStyle.definition.color &&
+      widgetStyle.definition.color.fixed;
```
rjimenezda: I think ramda / lodash have a thing for this pattern, but underscore doesn't 😞

matallo:
the thing is, I really grinded my gears to try to pull that value from elsewhere (imagine we change it in the future), but I couldn't find an elegant way to pass it around from builder :/

PR #560, opened 2017-07-10
time-series: add layer reference

matallo:
re: https://github.com/CartoDB/cartodb/issues/12446

- added layer reference to time-series, and torque widgets
- extended specs to make them pass

pageshot of untitled map carto 2017-07-01-1119 43
pageshot of untitled map carto 2017-07-01-1119 23

/cc @arianaescobar

### DIFF
--- a/src/widgets/time-series/content-view.js
+++ b/src/widgets/time-series/content-view.js
@@ -5,6 +5,8 @@ var contentTemplate = require('./content.tpl');
 var HistogramView = require('./histogram-view');
 var TimeSeriesHeaderView = require('./time-series-header-view');
 var DropdownView = require('../dropdown/widget-dropdown-view');
+var layerColors = require('../../util/layer-colors');
+var analyses = require('../../data/analyses');
 
 /**
  * Widget content view for a time-series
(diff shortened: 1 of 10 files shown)


src/widgets/time-series/content-view.js
```diff
@@ -23,12 +25,24 @@ module.exports = cdb.core.View.extend({
     this.clearSubViews();
     this.$el.empty();
 
+    var sourceId = this._dataviewModel.get('source').id;
```
rubenmoya: Just a small comment, since this variables are only used in one place we could use them directly instead of declaring variables.

src/widgets/time-series/torque-content-view.js
```diff
@@ -22,12 +24,24 @@ module.exports = cdb.core.View.extend({
   render: function () {
     this.clearSubViews();
 
+    var sourceId = this._dataviewModel.get('source').id;
```
rubenmoya: Same here.

rubenmoya:
Code looks good! Just a couple of comments but feel free to ignore them 👍

matallo:
🚀

PR #555, opened 2017-06-27
Time-series time aggregation

ivanmalagon:
Related to: https://github.com/CartoDB/cartodb/issues/12324

____________
cartodb PR: https://github.com/CartoDB/cartodb/pull/12347
cartodb.js PR: https://github.com/CartoDB/cartodb.js/pull/1685
deep-insights.js PR: https://github.com/CartoDB/deep-insights.js/pull/555

PR #562, opened 2017-07-18
time-series: add timezone selector to timeseries histogram

matallo:
re: https://github.com/CartoDB/cartodb/issues/12088

~~this will eventually point to master, when https://github.com/CartoDB/deep-insights.js/pull/555 is merged~~

---

cartodb PR: https://github.com/CartoDB/cartodb/pull/12464
deep-insights.js PR: https://github.com/CartoDB/deep-insights.js/pull/562
cartodb.js PR: https://github.com/CartoDB/cartodb.js/pull/1695

---

- [x] add support to local timezone toggle
- [x] tests

screen shot 2017-07-18 at 12 13 26

**acceptance**

when toggling the local timezone switch the timestamps in the chart should convert to local timezone, by default they should be in UTC, or in case another timezone is set, those timestamps should be in that timezone

the local timezone is not persisted, so whenever the visitor refresh the page the local timezone is reset to disabled


src/widgets/auto-style/auto-styler.js
```diff
@@ -1,107 +1,12 @@
 var cdb = require('cartodb.js');
 var CategoryColors = require('./category-colors');
-var AutoStyler = cdb.core.Model.extend({
+
+module.exports = cdb.core.Model.extend({
```
ivanmalagon: Why are there changes in auto-style in this PR? Does it come from a merge?

src/formatter.js
```diff
@@ -90,14 +91,16 @@ format.formatValue = function (value) {
   return value;
 };
 
-format.timestampFactory = function (aggregation) {
-  return function (timestamp) {
+format.timestampFactory = function (aggregation, offset) {
+  return function (timestamp, localTimezone) {
```
ivanmalagon: The original intent with this returned function was to provide a common way to `chart.js` to call the formatter, whether it was numeric or timestamp.

This comment continues at `chart.js`

src/widgets/histogram/chart.js
```diff
@@ -96,7 +96,7 @@ module.exports = cdb.core.View.extend({
 
     this.formatter = formatter.formatNumber;
     if (this._isDateTimeSeries()) {
-      this.formatter = formatter.timestampFactory(this._dataviewModel.get('aggregation'));
+      this.formatter = formatter.timestampFactory(this._dataviewModel.get('aggregation'), this._dataviewModel.get('offset'));
```
ivanmalagon: If everytime the timezone changes, we re-render the chart, we can pass that value to this function and then saving the `get('local_timezone')` in every call, exposing the same call to both numeric and date values.

Indeed, when we call `this.formatter` for numeric values, the parameter `local_timezone` has no sense.

But perhaps, it cannot be done if `initialize` is not called every time the timezone changes.

src/widgets/time-series/torque-time-info-view.js
```diff
@@ -18,32 +13,33 @@ module.exports = cdb.core.View.extend({
...
-        number: FORMATTER_TYPES['number'](scale(step))
+        time: timeFormatter(scale(this._torqueLayerModel.get('step')))
       });
     } else if (columnType === 'date' && !isNaN(time && time.getTime())) {
+      timeFormatter = formatter.timestampFactory(this._dataviewModel.get('aggregation'), this._dataviewModel.get('offset'));
+
       html = template({
-        time: FORMATTER_TYPES['time'](time),
-        date: FORMATTER_TYPES['date'](time)
+        time: timeFormatter(moment(time).unix(), this._timeSeriesModel.get('local_timezone'))
```
ivanmalagon: Perhaps this is an impediment for the previous comments. IDK

matallo:
retest this, please

src/widgets/dropdown/template.tpl
```diff
@@ -16,15 +16,31 @@
     </li>
   <% } %>
 
+  <% if (flags.localTimezone) { %>
+  <li class="CDB-Dropdown-item">
+    <button class="CDB-Dropdown-link js-toggleLocalTimezone">
+      Local Time Zone
```
rjimenezda: At some point we should add I18N to deep-insights 😞

matallo:
200w_d

PR #564, opened 2017-07-24
Add .CDB-Dashboard-hideMobile class

rubenmoya:
Related to: https://github.com/CartoDB/support/issues/865

The problem was the menu was displayed but with `opacity: 0;`, and it sits on top of the first checkbox if the Layer Selector option is enabled.

The menu has a class called `CDB-Dashboard-hideMobile` which doesn't exist, I just added it to the `max-width: 759px` media-query.


rubenmoya:
cc @matallo

matallo:
🚀

PR #570, opened 2017-08-08
Remove inline comments and transform variables into CSS for auto-style

xavijam:
Related ticket: https://github.com/CartoDB/cartodb/issues/12574

cc @rochoa @rubenmoya


package.json
```diff
@@ -35,7 +35,7 @@
     "node-polyglot": "^2.2.2",
     "perfect-scrollbar": "git://github.com/CartoDB/perfect-scrollbar.git#master",
     "postcss": "5.0.19",
-     "postcss-scss": "1.0.2",
```
rochoa: 😂

src/widgets/auto-style/style-utils.js
```diff
@@ -1,21 +1,26 @@
 var _ = require('underscore');
 var postcss = require('postcss');
+var stripInlineComments = require('postcss-strip-inline-comments');
+var SCSSsyntax = require('postcss-scss');
```
IagoLast: After a BIG research trying to find why this builds was failing I finally found that this is the guilty line!

For some reason when postcss-scss is included, jasmine is unable to find the tests breaking the build. (I thought it was something related to the CI speed improvement x.x )

I dont know the exact purpose of the PR but I see 3 options:

1. Try to use another library
2. Change the test runner
3. Try to find whats happening (trust me, this is the hardest!)

xavijam: You are right, looks like the last version is not working properly with our current jasmine thing. I'm using the same one from Grainstore (https://github.com/CartoDB/grainstore/blob/master/package.json#L26) and seems like it is alive again... let's see.

xavijam:
VAAAAMOS @IagoLast, thanks for your help mate 💃

src/widgets/auto-style/style-utils.js
```diff
@@ -1,21 +1,26 @@
...
 
 var OUTLINE_ATTRS = ['line-color', 'line-opacity'];
 
 function isPropertyIncluded (cartocss, attr) {
-  var cssTree = postcss().process(cartocss);
+  var cssTree = postcss()
```
IagoLast: This code is duplicated, what about creating a function?

```javascript
function createCssTree(cartocss){

return postcss()
        .use(stripInlineComments)
        .process(cartocss, { syntax: SCSSsyntax} );


```

src/widgets/auto-style/style-utils.js
IagoLast: should we open a ticket in the postcss-css repo?

xavijam:
All happy with it?

IagoLast:
LGTM

PR #579, opened 2017-08-28
Force resize when a widget is added/removed and there is a time-series

Jesus89:
The chart inside an histogram (time-series) is forced to resize when a widget is added or removed:

force-resize-widgets

I have added the public method "forceResize" to the dashboard, to enable its use from the Builder and also to avoid window.bind('resize') in the chart.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-deep-insights.js",
-  "version": "0.0.3",
+  "version": "0.0.3-12655",
   "description": "CARTO Deep Insights JS library",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-deep-insights.js",
-  "version": "0.0.3-12655",
+  "version": "0.0.3",
   "description": "CARTO Deep Insights JS library",
   "repository": {
     "type": "git",
--- a/spec/dashboard-below-map-view.spec.js
+++ b/spec/dashboard-below-map-view.spec.js
@@ -9,7 +9,11 @@ var createFakeTimeSeriesWidgetModel = function (attrs) {
     type: 'time-series'
   });
 
-  return new Backbone.Model(attrs);
+  var Widget = Backbone.Model.extend({
+    forceResize: function () {}
+  });
+
+  return new Widget(attrs);
 };
 
 var createFakeDataviewModel = function (attrs) {
@@ -67,11 +71,13 @@ describe('dashboard-below-map-view', function () {
     });
 
     describe('_onWidgetsChange', function () {
-      it('should trigger forceResize event', function () {
-        spyOn(this.timeSeriesWidgetModelFake, 'trigger');
-        this.view._onWidgetsChange();
+      it('should call forceResize event on each widget', function () {
+        this.widgetsCollection.each(function (widget) {
+          spyOn(widget, 'forceResize');
+          this.view._onWidgetsChange();
 
-        expect(this.timeSeriesWidgetModelFake.trigger).toHaveBeenCalledWith('forceResize');
+          expect(widget.forceResize).toHaveBeenCalled();
+        }, this);
       });
 
       it('should call _toggleVisiblity function', function () {
--- a/spec/dashboard-below-map-view.spec.js
+++ b/spec/dashboard-below-map-view.spec.js
@@ -60,10 +60,10 @@ describe('dashboard-below-map-view', function () {
 
   describe('when a time-series widget is added', function () {
     beforeEach(function () {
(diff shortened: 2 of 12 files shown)


src/widgets/histogram/chart.js
```diff
@@ -586,7 +586,7 @@ module.exports = cdb.core.View.extend({
   _setupDimensions: function () {
     this._setupScales();
     this._setupRanges();
-    this._onWindowResize();
+    this.onWindowResize();
```
xavijam: Why have we changed this method without the `_`?

src/dashboard-below-map-view.js
```diff
@@ -71,6 +71,11 @@ module.exports = cdb.core.View.extend({
   },
 
   _onWidgetsChange: function () {
+    this._widgets.each(function (widget) {
+      if (widget.get('type') === 'time-series') {
```
rubenmoya: I'd use a constant here, since we're using that string multiple times in the file

src/widgets/histogram/chart.js
Jesus89: Now this method is used outside chart.js, so it has become public

src/dashboard-below-map-view.js
Jesus89: Yep

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cartodb-deep-insights.js",
-  "version": "0.0.3",
+  "version": "0.0.3-12655",
```
nobuti: This is unnecessary.

spec/widgets/time-series/histogram-view.spec.js
```diff
@@ -131,6 +136,19 @@ describe('widgets/time-series/histogram-view', function () {
     });
   });
 
+  describe('_onForceResize', function () {
+    it('should call _chartView.onWindowResize function', function () {
+      this.view._chartView = {
```
nobuti: I think you can save a few of lines with `this.view._chartView = jasmine.createSpyObj('chartView', ['onWindowResize']);`.

src/dashboard-below-map-view.js
```diff
@@ -71,7 +73,11 @@ module.exports = cdb.core.View.extend({
   },
 
   _onWidgetsChange: function () {
+    this._widgets.each(function (widget) {
```
nobuti: we can save iterations if we use the `some` iterator returning false after the trigger.

Jesus89: Thanks! Btw, I'm still working on this.

spec/widgets/time-series/histogram-view.spec.js
Jesus89: Thanks! :+1:

src/dashboard-below-map-view.js
nobuti: Oh, I picked up because it's in code review. Sorry about that.

Jesus89: Sorry, I have continued the development because this related issue: https://github.com/CartoDB/cartodb/issues/12124

spec/dashboard-below-map-view.spec.js
```diff
@@ -56,15 +60,33 @@ describe('dashboard-below-map-view', function () {
 
   describe('when a time-series widget is added', function () {
     beforeEach(function () {
-      var timeSeriesWidgetModelFake = createFakeTimeSeriesWidgetModel();
-      timeSeriesWidgetModelFake.dataviewModel = createFakeDataviewModel();
-      timeSeriesWidgetModelFake.dataviewModel.layer = createFakeDataviewLayer();
-      this.widgetsCollection.add(timeSeriesWidgetModelFake);
+      this.timeSeriesWidgetModelFake = createFakeTimeSeriesWidgetModel();
```
xavijam: Why do we create this variable inside `this`?

xavijam: If we are not going to use it out of this context, I'd not create it.

src/dashboard-view.js
```diff
@@ -18,6 +19,8 @@ module.exports = cdb.core.View.extend({
     // TODO parent context requires some markup to be present already, but NOT the other views
     this.el.classList.add(this.className);
     this.$el.html(template());
+
+    $(window).bind('resize', this._onWindowResize.bind(this));
```
xavijam: I feel the unbind will not work unless:

```js
this._onWindowResize = this._onWindowResize.bind(this);
$(window).bind('resize', this._onWindowResize);
…

$(window).unbind('resize', this._onWindowResize);
```

spec/dashboard-below-map-view.spec.js
Jesus89: That's correct! This remains from a previous change

PR #582, opened 2017-08-31
Add tooltips to show bucket data in time-series

Jesus89:
Related to: https://github.com/CartoDB/cartodb/issues/11650

bucket-tooltip


src/widgets/constants.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
xavijam: 😱

src/widgets/time-series/histogram-view.js
```diff
@@ -131,6 +136,27 @@ module.exports = cdb.core.View.extend({
     this._chartView.forceResize();
   },
 
+  _clearTooltip: function () {
+    this.$('.js-tooltip').stop().hide();
+  },
+
+  _onValueHover: function (info) {
```
xavijam: Q: Don't we have already a tooltip view for managing this?

src/widgets/constants.js
xavijam: Following @rubenmoya constants.. 🆗 , but there are a bunch of them we could add here 😬  (it is not necessary to add them now!)

spec/widgets/time-series/histogram-view.spec.js
```diff
@@ -107,6 +107,18 @@ describe('widgets/time-series/histogram-view', function () {
...
+      chartMock.show = function () {};
+      spyOn(this.view, '_instantiateChartView').and.returnValue(chartMock);
+
+      this.view._createHistogramView();
+
+      var html = this.view.$el.html();
+
+      expect(html.indexOf('CDB-Widget-tooltip') > 0).toBe(true);
+    });
   });
```
xavijam: When user make hover, we should check that the tip appears

src/widgets/constants.js
rubenmoya: About the constants, I think we should have only one `constants.js` file, in the root folder (`src`), and use it only for constants which will be shared across the project (map types, layer types, etc). If the constant is only used in one file, I suggest to keep it on that file. If the `constants.js` files starts getting big, we could always split it in multiple files in a `/constants` folder.

src/widgets/time-series/histogram-view.js
```diff
@@ -1,21 +1,24 @@
...
+var C = require('../constants');
 
 /**
  * Time-series histogram view.
  */
 module.exports = cdb.core.View.extend({
+  className: 'CDB-Chart--histogram',
+
   defaults: {
+    offsetLeft: -2,
```
nobuti: why is -2? Does offsetLeft refer to the histogram?

src/widgets/time-series/histogram-view.js
nobuti: As @xavijam said, we already have a tooltip view for widgets. But it's not ready for this because it depends on a target. Maybe it's worth that we check what it needs to be used here too, and clean the code repetition we have right now.

nobuti:
Done!

src/widgets/widget-tooltip-view.js
```diff
@@ -1,58 +1,74 @@
...
   },
 
-  show: function () {
+  _setValue: function (target) {
+    var $target = $(target);
+    var value = $target.attr(this.options.attribute);
+    this.$el.html(value);
+  },
+
+  show: function (e) {
```
rubenmoya: `e`, that could mean `error` too, I think we should use full words in function arguments / variables.

src/constants.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  TOOLTIP_OFFSET_Y: 10
```
rubenmoya: If that's only used in widgets maybe we should "namespace" it: 

```
module.exports = {
  WIDGETS: {
    TOOLTIP_OFFSET_Y: 10
  }
}
```

so it get's easier to refactor later if we have tons of constants:

```
module.exports = {
  WIDGETS: require('./widgets')
}
```

src/widgets/histogram/chart.js
```diff
@@ -1499,6 +1500,9 @@ module.exports = cdb.core.View.extend({
       })
       .attr('y', self.chartHeight())
       .attr('height', 0)
+      .attr('data-tooltip', function (d) {
```
rubenmoya: Not a big fan of one-letter function arguments, could we use `data` or `tooltipData` or something more explicit?

nobuti: This is really a convention from d3, I see your point but two reasons why I won't fix:
- The "parent" of that `d` attribute is `data` so it would be really confusing. Also, that `d` depends on the use you are doing of that value. Some times it would be related to the tooltip, other to axis, labels or something else.
- There are a lot of places in the code to be changed, names to get addressed... I get you but it's out of scope.

nobuti:
CR taken

src/widgets/histogram/chart.js
rubenmoya: 👍

rubenmoya:
To acceptance!

ivanmalagon:
retest this please

src/widgets/histogram/chart.js
```diff
@@ -1499,6 +1501,9 @@ module.exports = cdb.core.View.extend({
...
+        return self._tooltipFormatter(d.freq);
+      })
```
ivanmalagon: `this.formatter` depends on the column type. For tooltips, we always need to format as a number.

PR #583, opened 2017-08-31
Trigger indexes change event even if they don't change

rubenmoya:
Related to: https://github.com/CartoDB/cartodb/issues/12661

The handlers where getting stuck in the middle of the buckets, debugging I found that if you moved the handler to the first half of the next bucket, it should move where it was previously, which means the same indexes, so the backbone model didn't trigger any changes which lead to the function which calculates the handler position not being called.

The proposed solution is to make the changes silent, and the trigger the event manually.


nobuti:
Works like a charm ✨

PR #585, opened 2017-09-05
Escape HTML content in widgets source

Jesus89:
Related to: https://github.com/CartoDB/cartodb/issues/12322

Changes:
- Escape HTML in source layerName for all the widgets. This change prevents XSS (https://github.com/CartoDB/bug-bounty/issues/152)
- Refresh widgets when a Layer's title changes, to update its sources

### DIFF
--- a/src/widgets/category/title/search-title-view.js
+++ b/src/widgets/category/title/search-title-view.js
@@ -75,6 +75,7 @@ module.exports = cdb.core.View.extend({
     this.add_related_model(this.dataviewModel.filter);
 
     this.dataviewModel.layer.bind('change:visible change:cartocss', this.render, this);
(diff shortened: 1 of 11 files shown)


src/widgets/category/title/search-title-view.js
```diff
@@ -48,7 +48,7 @@ module.exports = cdb.core.View.extend({
         sourceType: analyses.title(sourceType),
         showSource: this.model.get('show_source') && letter !== '',
         sourceColor: sourceColor,
-        layerName: layerName,
+        layerName: _.escape(layerName),
```
rubenmoya: Should we wrap this in a util function like we do in cartodb? so in the future if we change how we sanitise thins we will have to change it only in one place

PR #587, opened 2017-09-11
Improve lint task

Jesus89:
Changes:
- Use eslint grunt-task instead of semistandard binary
- Update semistandard package to 11.0.0
- Remove semistandard config from package.json
- Fix .eslintignore file
- Update .eslintrc.json file
- Fix Gruntfile lint
- Fix various rules lint
- Add eslint to hound

NOTE: rule "no-mixed-operators" is escaped in eslint.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,24 @@
+{
+  "extends": "semistandard",
+  "env": {
+    "es6": true,
+    "browser": true,
+    "jasmine": true
+  },
+  "globals": {
+  },
+  "rules": {
+    "key-spacing": "off",
+    "comma-dangle": "off",
+    "spaced-comment": "off",
+    "padded-blocks": "off",
+    "comma-spacing": "off",
+    "semi": "off",
+    "indent": "off",
+    "no-undef": "off",
+    "no-empty-label": "off",
+    "space-before-keywords": "off",
+    "space-after-keywords": "off",
+    "space-return-throw-case": "off"
+  }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,10 +8,6 @@
   "globals": {
   },
   "rules": {
-    "comma-dangle": "off",
-    "spaced-comment": "off",
-    "padded-blocks": "off",
-    "comma-spacing": "off",
     "semi": "off",
     "indent": "off",
     "no-undef": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,11 +11,6 @@
     "Clipboard": true
   },
   "rules": {
-    "space-before-keywords": "off",
-    "space-after-keywords": "off",
-    "space-return-throw-case": "off",
-    "no-mixed-operators": "off",
-    "no-use-before-define": "off",
-    "no-multiple-empty-lines": "off"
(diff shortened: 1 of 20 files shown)


alonsogarciapablo:
I trust you guys 300% 👍 . @IagoLast is the ESLint fan/master/evangelist and might have something to say...

.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "semistandard",
+  "env": {
+    "es6": true,
+    "browser": true,
+    "jasmine": true
+  },
+  "globals": {
```
IagoLast: I dont know the code very well but could we use requires instead?

package.json
```diff
@@ -21,7 +21,8 @@
     "Antonio Santiago Pérez <asantiago@cartodb.com>",
     "Ivan Malagon <ivan@carto.com>",
     "Ruben Moya <ruben@carto.com>",
-    "Carlos Matallín <matallo@carto.com>"
+    "Carlos Matallín <matallo@carto.com>",
+    "Jesús Arroyo Torrens <jarroyo@cartodb.com>"
   ],
   "license": "BSD-3-Clause",
   "dependencies": {
```
IagoLast: IMO lint things should be part of the dev-deps

.eslintrc.json
Jesus89: $ and Clipboard are added because of the styleguide.js file

package.json
```diff
@@ -77,32 +80,9 @@
...
-      "/grunt-tasks",
-      "/styleguide",
-      "/themes",
-      "/vendor"
-    ]
-  },
   "scripts": {
-    "test": "grunt test"
+    "test": "grunt test",
+    "lint": "./node_modules/.bin/eslint .",
```
IagoLast: Since [npm already resolves it](https://docs.npmjs.com/misc/scripts#path)  I think we can use eslint without specifying the path

Jesus89: Great!

package.json
Jesus89: What do you mean? eslint and grunt-eslint are in devDependencies.

.eslintrc.json
IagoLast: what about [specifying-globals](https://eslint.org/docs/user-guide/configuring#specifying-globals) in `styleguide.js` file and removing them from the config?

.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
```
IagoLast: ### Personal and subjective comment alert!

I prefer the eslintrc to be a javascript file.

 - Is smaller
 - Allows comments
 - No json shitty errors because commas!

.eslintrc.json
Jesus89: It can be done this way

package.json
IagoLast: ouch you are right. sorry I was looking at the diff collapsed

.eslintrc.json
Jesus89: We had the same issue between .eslintrc and .eslintrc.json :laughing:

@CartoDB/frontend any comment?

rubenmoya:
Working as expected, the next step would be get rid of grunt!

PR #589, opened 2017-09-14
Glitches for histogram charts

nobuti:
This PR improves the histograms rendering. It's related to https://github.com/CartoDB/cartodb/issues/12217. Right now, a histogram has a few issues:
- In Safari, all bars are rendered without spacing, the worst scenario occurs for time-series.
- When a user applies a zoom to the browser, the spacing is not consistent between bars. This is for all browsers.

### DIFF
--- a/themes/scss/widgets/_default.scss
+++ b/themes/scss/widgets/_default.scss
@@ -77,6 +77,10 @@
 
 .CDB-Chart--histogram {
   margin-top: $baseSize * 3;
+
+  @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
+    transform: rotateZ(-0.001deg);
+  }
 }
 
 .CDB-Widget-content--histogram,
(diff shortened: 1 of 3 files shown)


themes/scss/widgets/_default.scss
```diff
@@ -77,6 +77,10 @@
 
 .CDB-Chart--histogram {
   margin-top: $baseSize * 3;
+
+  @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
```
nobuti: This targets only IE10+.

themes/scss/widgets/_default.scss
```diff
@@ -77,6 +77,10 @@
...
+    transform: rotateZ(-0.001deg);
```
rubenmoya: So hacky 😂

nobuti: Indeed 😞

PR #595, opened 2017-09-21
Added DO multiple locale.

nobuti:
This PR fixes https://my.trackjs.com/details/0bf166fff2074fd9b0dc7befc54dcbd1


ivanmalagon:
The hotfix looks good to me.

The only note here is that this will happen again when a new analysis is added to camshaft reference. I don't think is nothing too dangerous (it doesn't break, only shows the untranslated string in the UI and throws an error to the console).

nobuti:
Yes, but even it's not serious, we need to figure out how to improve it to avoid this in the future.

Issue #574, opened 2017-08-11
Improve widgets error state for supporting several error types


xavijam:
When there is an error in the Dataviews request, widgets can be rendered. We display an error saying "There has been an error" (or something like that), and a refresh button (is it useful?).

From now on we should be more concrete about the problem. We should catalog those by:

- Data error: any dataview request returned an error and we should display it.
- Limits error: the user has reached its platform limits, dataview request will fail but with a specific error code or message, so we should decide if the error message should change or not (**let's not forget this will be public in the embed page**).

## Design

![d60763a6-3a63-11e7-8d97-120b649726e3](https://user-images.githubusercontent.com/132146/29208324-e65d7eb0-7e89-11e7-8937-a04a80a645cf.png)

*Related ticket: https://github.com/CartoDB/cartodb-platform/issues/3279

rubenmoya:
Hi! I've seen in this issue https://github.com/CartoDB/cartodb/issues/12751 the widget error text has changed, right now, for the limits engine is sending this error:

```
{
  type: 'limit',
  message: 'You are over platform\'s limits. Please contact us to know more details',
  subtype: 'datasource'
}
```

Should we separate both errors, timeout (poor connectivity, error with the server) and limits (database/tiler timeout), or keep the same error for both of them?

Also, I've realised, since we're going to have more feedback about the errors, is it still necessary the design with only one button or can we use the last one in this design for that case?

![30542755-df01e812-9c80-11e7-9e45-12eb7727afe2](https://user-images.githubusercontent.com/16247628/30542755-df01e812-9c80-11e7-9e45-12eb7727afe2.png)

cc @CartoDB/design

arianaescobar:
IMO, we should separate them. We might want to improve the copy for that limits error and prepare the visual. 

Regarding your last question: if we have specific information about the error (like limits or poor connectivity) I would show the full message to properly explain it. But if we don't, I recommend avoiding the repetition of elements. Right now, in the following case, the buttons already feel super repetitive, if we also add a generic message and repeat it all over, it will be too much. 

![30153886-6e0539c8-93b8-11e7-8952-24cae556e718](https://user-images.githubusercontent.com/16247628/30699445-11377908-9ee5-11e7-976a-1c3d169ddc13.png)

noguerol:
I totally agree, @CartoDB/design . New ticket for that, please!

rubenmoya:
[...]

PR #597, opened 2017-09-25
Improve errors in widgets

rubenmoya:
Related to: https://github.com/CartoDB/deep-insights.js/issues/574

screen shot 2017-09-25 at 12 08 19
screen shot 2017-09-25 at 12 09 42
screen shot 2017-09-25 at 12 17 36

## Acceptance steps
- [ ] Should show the timeout error for dataviewmodel and datamodel requests with status `429`
- [ ] With the error view active, should show the widget normal view if the next request is success
- [ ] The refresh button should work for dataviewmodel and datamodel requests
- [ ] Should show the _no data available_ error if there are no points in the bbox
- [ ] WIth the _no data available_ should show the widget normal view if there are points in the bbox
- [ ] Should refresh the data when moving around the map before and after an error


xavijam:
Tiene pintote.

spec/widgets/widget-view-factory.spec.js
```diff
@@ -55,6 +55,7 @@ describe('widgets/widget-view-factory', function () {
       this.widget = new cdb.core.Model({}, {
         layer: this.layer
       });
+      this.widget.dataviewModel = {};
```
xavijam: Not need anything else?

src/widgets/formula/placeholder.tpl
```diff
@@ -0,0 +1 @@
+<div class="CDB-Widget-listItem--fake"></div>
```
xavijam: Eing? Didn't we have one of these?

src/widgets/widget-error-view.js
```diff
@@ -1,49 +1,56 @@
...
-
-  show: function () {
-    this.$el.removeClass('is-hidden');
+    this._model.refresh();
   },
 
-  hide: function () {
-    this.$el.addClass('is-hidden');
+  _reset: function () {
+    this.$el.removeClass('CDB-Widget--error');
```
xavijam: Do you know you can remove several classes in the same instance?

src/widgets/widget-view.js
```diff
@@ -15,26 +24,71 @@ module.exports = cdb.core.View.extend({
 
   initialize: function () {
     this.listenTo(this.model, 'destroy', this.clean);
```
xavijam: I'd create:
```js
var dataviewModel = this.model.dataviewModel;
this.listenTo(dataviewModel, 'sync change:data', this._onDataChanged);
if (dataviewModel && dataviewModel._totals) {
  …
}
```

xavijam:
Ok, tested in staging and I've found a problem trying to refresh a histogram widget (refresh is undefined). Back to development!

spec/widgets/widget-view-factory.spec.js
rubenmoya: No, just needed because of `dataviewModel._totals` would be `undefined.totals` otherwise

rubenmoya:
retest this please

PR #615, opened 2017-10-25
Integrate extract-engine

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1837


spec/widgets/time-series/torque-histogram-view.spec.js
```diff
@@ -32,7 +32,7 @@ describe('widgets/time-series/torque-histogram-view', function () {
       start: 0,
       end: 1000
     }, {
-      vis: vis
+      engine: vis._engine
```
alonsogarciapablo: Should we expose a `getEngine` method?

Jesus89: This is required only in some tests, and it could change in the future.

PR #629, opened 2017-11-23
Dataview enable always true, no matter layer visibility 

xavijam:
Related ticket: CartoDB/support#1135


xavijam:
Tested with support viz.json and the `easy.html` example and working 👌 . Merging!

Issue #640, opened 2017-12-18
Missing hover in widgets


piensaenpixel:
Related - https://github.com/CartoDB/design/issues/1035

cc @rubenmoya @matallo

matallo:
whoa, many elements I didn't have in my radar, let me prioritize this for this next sprint

matallo:
in production! 🚀

piensaenpixel:
<3

piensaenpixel:
but one thing, if i change the color-widget to a custom color, the hover doesn't work :(

rubenmoya:
It seems like we broke it with this change: https://github.com/CartoDB/deep-insights.js/pull/641/files#diff-11ebf18b812ac90a72a6c99eff5f150eR1444, it was using the `currentFillColor` and we change it to a hardcoded one.

jesusbotella:
Yes, it is my fault :(

rubenmoya:
I did the CR, and someone else did the acceptance, and none of us noticed, don't blame yourself, these things happen, let's fix it and focus on what we've learned about widgets autostyling :)

matallo:
removing from board for better tracking

rubenmoya:
In production! 🚀

PR #641, opened 2017-12-18
Add missing hover and selected state colors to Widgets

jesusbotella:
Related to #640 

- Deep-insights.js https://github.com/CartoDB/deep-insights.js/pull/641
- CartoDB https://github.com/CartoDB/cartodb/pull/13275

----

This PR adds the missing hover and selected state colors to Widgets.
Based on https://user-images.githubusercontent.com/14546701/32537554-50cc906e-c463-11e7-9e20-5c11a5ce4a32.png

As I removed `tinycolor` dependency and it is not used anywhere else, do you want me to remove it from the project?

#### Category Widgets

- Check that `Bar Color Green with 16% black` color is applied when mouse is over any item
- Check that `Bar Color Dark` is applied when any item is selected

#### Histogram Widgets

- Check that `Bar Color Green with 16% black` color is applied when mouse is over any bar
- Check that `Bar Color Dark` is applied when there is a range selected

### DIFF
--- a/src/widgets/histogram/chart.js
+++ b/src/widgets/histogram/chart.js
@@ -3,13 +3,13 @@ var _ = require('underscore');
 var d3 = require('d3');
 var d3Interpolate = require('d3-interpolate');
 var cdb = require('cartodb.js');
-var tinycolor = require('tinycolor2');
 var formatter = require('../../formatter');
 var timestampHelper = require('../../util/timestamp-helper');
 var viewportUtils = require('../../viewport-utils');
 
-var FILTERED_COLOR = '#1181FB';
+var FILTERED_COLOR = '#2E3C43';
 var UNFILTERED_COLOR = 'rgba(0, 0, 0, 0.06)';
+var HOVER_COLOR = '#82BB90';
 var TIP_RECT_HEIGHT = 17;
 var TIP_H_PADDING = 6;
 var TRIANGLE_SIDE = 14;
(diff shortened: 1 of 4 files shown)


jesusbotella:
@urbanphes I noticed that the `Other` category in the example has a solid background when disabled, and right now in the Builder is looks like this:

screen shot 2017-12-18 at 18 20 03

Do you want me to change it?

src/widgets/histogram/chart.js
```diff
@@ -3,13 +3,13 @@ var _ = require('underscore');
 var d3 = require('d3');
 var d3Interpolate = require('d3-interpolate');
 var cdb = require('cartodb.js');
-var tinycolor = require('tinycolor2');
```
rubenmoya: Is it used anywhere else in the project? If not we could remove it from the `package.json`

jesusbotella: I think it is not used anywhere else, I searched and there were only two occurrences: https://github.com/CartoDB/deep-insights.js/search?utf8=%E2%9C%93&q=tinycolor&type=

So I am going to remove it from package.json

urbanphes:
@jesusbotella sorry. We let "OTHER" bar with the patterns styles a production.

captura de pantalla 2017-12-21 a las 9 41 16

matallo:
then back to development!

PR #642, opened 2017-12-19
Improve node info in widget

rubenmoya:
Related to: https://github.com/CartoDB/cartodb/pull/13239#issuecomment-352463614
Apply some utility classes to center the node information vertically in the widget

### DIFF
--- a/src/widgets/category/title/search-title-template.tpl
+++ b/src/widgets/category/title/search-title-template.tpl
@@ -39,8 +39,8 @@
         </span>
 
         <% if (!isSourceType) { %>
-          <span class="CDB-Text CDB-Size-small u-lSpace--s" style="color: <%- sourceColor %>;">
-            <i class="CDB-IconFont CDB-IconFont-ray"></i>
+          <span class="CDB-Text CDB-Size-small u-lSpace--s u-flex" style="color: <%- sourceColor %>;">
+            <i class="CDB-IconFont CDB-IconFont-ray CDB-Size-small"></i>
           </span>
         <% } %>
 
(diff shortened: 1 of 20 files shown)


src/widgets/category/title/search-title-template.tpl
```diff
@@ -39,8 +39,8 @@
         </span>
 
         <% if (!isSourceType) { %>
-          <span class="CDB-Text CDB-Size-small u-lSpace--s" style="color: <%- sourceColor %>;">
-            <i class="CDB-IconFont CDB-IconFont-ray"></i>
+          <span class="CDB-Text CDB-Size-small u-lSpace--s u-flex" style="color: <%- sourceColor %>;">
+            <i class="CDB-IconFont CDB-IconFont-ray CDB-Size-small"></i>
```
matallo: Does it make sense using `CDB-Size-small` without `CDB-Text` in that same element?

rubenmoya: By [watching the classes](https://github.com/CartoDB/CartoAssets/blob/master/src/scss/cdb-components/_typography.scss), it makes sense (I mean, it works), but maybe semantically it doesn't make sense

matallo: I've had unintended consequences before using both classes separately then together, I mean they are not really utility classes but they affect each other's behaviour :/

That's why I would be concerned

rubenmoya: I tried to change it and I realized one thing, if I use CDB-Text it changes the font-family, so the icon disappear :joy:

Maybe it's just a naming problem? And it should be `CDB-Font` instead of `CDB-Text` so we have `CDB-Font / CDB-Iconfont` which changes `font-family` and then `CDB-Size`?

matallo: Ha! That's what I'm talking about, if it looks ok with the current changes, then 👍

matallo:
retest this, please

PR #649, opened 2017-12-21
Display decades in time series widget

ivanmalagon:
Related to https://github.com/CartoDB/support/issues/1071
----

This PR shows decade labels in time series widget.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "backbone": "1.2.3",
     "cartocolor": "4.0.0",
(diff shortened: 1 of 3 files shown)


Algunenano:
Can you add century and millennium for the same price?

ivanmalagon:
I talked with @alonsogarciapablo yesterday and from the UX perspective doesn't make much sense right now.
If we encounter any future issue that can be resolved with those aggregations we can refer to these PRs to add them to the UI.

